PR #27, opened 2021-06-08
Add internationalisation (`react-i18next`)

gpxl-dev:
Approach is explained in `i18n.md` within this PR.

### Discussion points:

1. Do we need/want the localStorage cache?
2. Does the proposed structure for the resource files make sense?

*Note there are a couple of other minor changes in this PR: removing package-lock and updating the favicon/manifest. The latter changes are still temporary pending designer/marketing input but probably closer to where we want to be for now. If it's contentious let me know and I'll make this PR strictly i18n only.*


dmosites (approved):
LGTM!

dmosites (approved):

PR #45, opened 2021-06-11
Tailwind css

gpxl-dev:
This adds [tailwind css](https://tailwindcss.com/) to the project. Best bet is to read the docs if you want to understand how to use it, but I've also created a few example components, most notably the token selector. For now pretty much everything is subject to change for designs, so please consider any actual use of tailwind in this PR as **for example only**.

As you will see from the docs, tailwind generally recommends extracting components (e.g. `Button.tsx` for reducing repeated classes, but where this is considered overkill, you can also write general rules by using the `@apply` at rule. See `index.css` for an example of where I've done thins.

I make heavy use of the `classNames` package when using tailwind, not least for laying things out.

Suggest using this PR to acquaint yourself with Tailwind if you haven't already, then we can decide whether this is the route we'd like to go down for styling.

P.S. just implementing a couple of components threw up a few questions (folder structure, for example) - if there's anything I've done that is contentious it's probably worth brining up now as I had to make a bunch of decisions.

### Changes

- Added tailwind as per [guide here](https://tailwindcss.com/docs/guides/create-react-app)
- Added vscode tailwind plugin to repo recommendations
- Added tailwind CSS at rule definitions to vscode config ([via Custom Data for CSS Language Service](https://github.com/Microsoft/vscode-css-languageservice/blob/main/docs/customData.md))
- Added [`classnames`](https://www.npmjs.com/package/classnames) package
- Added dark mode switch component `Components/DarkModeSwitch.tsx`
  - When explicitly enabled/disabled, the `.dark` class is added to/removed from the document element, and the theme (`"light"` or `"dark"` is persted in localStorage)
  - On page load, the localStorage preference is used if it exists, otherwise the OS theme is used.

### Temporary changes (pending design finalisation)

- All styling & components
- Added `react-icons` (although I really like this, but if we have designers to make us custom icons, then that's awesome).
- Added a range of blues for AirSwap colour palette (used [maketintsandshades.com](https://maketintsandshades.com/#2b72ff)), named `primary`.
  - [Tailwind Reference](https://tailwindcss.com/docs/customizing-colors#curating-colors)


dmosites (approved):
LGTM

PR #47, opened 2021-06-14
Add matomo as package.

gregplumbly:
Added one sample event to track.


gpxl-dev:
Code looks good. Two notes:

1. When I ran it, `matomo.js` was blocked by my browser - I was using Brave and it aggressively blocks a ton of stuff. It would be good to know if matomo have had any discussions with Brave to whitelist their stuff, given they are privacy centric. If not it would be interesting to know if there _are_ any analytics providers that Brave has whitelisted?

2. After setting "shields down" for the site, I got `Error in setSecureCookie: You cannot use `Secure` on http.` in the console. This is because by default the CRA dev server is http. Obviously it won't be an issue in production, but if there is anyone like me who doesn't like red text in the console, you can get around this by setting up https localhost:

```bash
brew install mkcert
cd /path/to/airswap-web/
mkcert localhost
```

Then create `.env.local` with the following contents:

```env
HTTPS=true
SSL_CRT_FILE=localhost.pem
SSL_KEY_FILE=localhost-key.pem
```

& `yarn start`

Probably want to update gitignore to ignore `.pem` files too.

gpxl-dev (approved):
Looks great from a code standpoint. See comment re:brave.

PR #54, opened 2021-06-15
Storybook

gpxl-dev:
This PR adds storybook with minimal custom config.

[Storybook docs are here](https://storybook.js.org/docs/react/get-started/introduction)

To run storybook: `yarn run storybook`

### Notes

- Added [tailwind dark mode addon](https://storybook.js.org/addons/storybook-tailwind-dark-mode/) which adds a moon icon in the toolbar that toggles the dark mode class for components. Currently haven't worked out how to change the background behind the components though so this will need some additional work - for now you can change the background  with the background button two to the left of the dark mode one. (PS - for the button story I've implemented, only the `neutral` intent button is affected by dark mode).
- I've added a story for `Button` just to show how they're written, but you may need to reference docs for more complex stories. Note the JSDoc comments on prop types are pulled into the "Docs" section in storybook, so documenting props will be good practice
- When specifying component titles, if you use a folder structure (`title: "components/Button",`) the components are listed in that structure. For now I suggest we just reflect the file system structure.
- Down the line we can also add documentation for non-components - check out mdx stories in the docs.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,39 @@
+const path = require("path");
+module.exports = {
+  // ref: https://storybook.js.org/docs/ember/configure/typescript#mainjs-configuration
+  typescript: {
+    check: false,
+    checkOptions: {},
+    reactDocgen: "react-docgen-typescript",
+    reactDocgenTypescriptOptions: {
+      shouldExtractLiteralValuesFromEnum: true,
+      propFilter: (prop) =>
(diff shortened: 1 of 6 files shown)


gpxl-dev:
Reason for `resolve` in package.json: https://github.com/storybookjs/storybook/issues/13183

codyenokida (approved):
Looks good, definitely can use this as a template for dev.

PR #59, opened 2021-06-17
Cooodes working issue 22 (Save Token Approval through localStorage and Redux Slice)

codyenokida:
A couple of changes that I made to save token approval state for individual tokens. 

1. Added a TokenApprovalState key/value pair to save the necessary token and their respective approval state. Value enums are "incomplete" | "pending" | "complete". This state is saved as an Object (key/value pair), with default values of {"": "incomplete"} (keys: tokenAddress, value: enum ("incomplete" | "pending" | "complete"))
2. Created a new Redux Slice for tokenApprovals. 
3. Whenever the state is changed, localStorage is updated. 
4. added useEffect to enable localStorage change to the Redux State on initial load.

I don't know if I made this way too complicated or not, one of my first times using redux and localStorage!


gpxl-dev (dismissed):
Not too far off I don't think.

A few things though:

1. Not sure if you know this or not (sorry if you do!) , but the amount of your tokens that a contract is allowed to spend on your behalf is called the `allowance` - this is what is set when calling `approve`. Currently the `balances` feature fetches and tracks changes to `allowance` for all active tokens. We have a temporary UI that prints out this info in the table beneath the swap widget. As such, this is the authoritative source of truth that we should use to determine whether or not to disable the approve button. I actually [made this change on the routing branch yesterday](https://github.com/airswap/airswap-web/compare/routing#diff-af79b6d36650c1d0a7f6c8ae828ea35751ee33de080186ef9584e46c7ab0768aR56) - bit of a stealth change, sorry.
2. In terms of changes to the redux state - I actually think this would sit better in either `allowances` or `transactions` rather than it's own state. Probably the latter, but you would need to update that state to be able to deal with both submitted orders and approval transactions.
3. Your current implementation appears to assume success - if I cancel the approval request, the approve button is disabled until I clear my localStorage, even though my `allowance` is still 0. It is actually worth noting that (in terms of the button state specifically), because we are already tracking allowances, we don't technically need to care about whether the approval transaction is successful or not, we just need to know that it is not still pending, because provided we update the button's disabled state (or visibility in my implementation) from the `allowances` state, we only need to know to toggle `loading` or not. On the other hand, though, we will probably want to tell the user it failed if it does, so it's worth differentiating.
4. With `2` and `3` in mind, I would take a look at the way Don did the `transactions` feature, and see if you can extend that slightly to allow for multiple types of transaction. One idea that I haven't thought through fully is changing to the following in `transactionsSlice`:
    image
5. We have some state (list of `activeTokens`) being persisted to localStorage already - we did this via `store.subscribe` in `metaDataSubscriber`. Updating in a thunk is also fine I think, but I don't think it was possible with `activeTokens`, and I feel like it might be nice to keep all localStorage stuff in one place, so you could consider moving the `localStorage` updates to here. The other thing to consider is that a pending transaction may not be pending on the next load (e.g. if a user leaves it pending and closes the tab for an hour and comes back later). ~~I'm not actually sure how to deal with that... 🤔 . I think we should be able to make some good guesses about whether or not the approval has completed (`allowance` will be non zero) so should be able to clear it with some confidence in the event of success, but if the tx fails we will not know about it. My only thought is to add a timestamp, then assume that after some sensible amount of time the tx has failed if it's still pending and re-enable the button. This is not perfect as it risks concurrent approval requests and hence wasted gas, but the alternative is to always have the button enabled which risks that anyway.~~ (**EDIT**: see below)
    


Also, just in case you don't have them - download the redux dev tools which are pretty cool in general and allow you to see the current state (below), and rewind through actions and watch how it changes, etc.
image

gpxl-dev:
re: 5. above - it's probably possible to look up a transaction once it has been mined or reverted. I was off in the land of mempools but by that stage I think there should be a record on chain, so actually there is most likely no ambiguity there?

I've never done it, so @dmosites can probably comment with authority, but a bit of googling should surface the answer, I'm sure!

gpxl-dev:
Notes after our call, as succinct as I can be!

- Update `transactionsSlice` to accept approvals transactions (follows that you also need to update `transactionActions`
- Add `{ dispatch }` param to the `approve` thunk in `orderSlice` (reference typing from `balancesSlice:51-61` if you want the correct types to be pulled through).
- Dispatch the correct action to add the approval transaction to `transactions` in the state.
- set the `loading` prop based on whether `transactions` contains a pending approval transaction for the sender token (filter the `transactions` state which you can either do in a selector e.g. `selectApprovalTxForToken(addr: string)` or select all transactions and filter it in the orders component itself.
- On page load, set state from local storage and look for any transactions that were pending. Use `getTransaction` to find if they have succeeded or been reverted (in which case just update the state but do nothing else), or if they are still pending. If you get a `TransactionResponse` back, great - wait for `wait` to resolve and update the state, if you get `null` then probably poll `getTransaction` at a sensible interval.

codyenokida:
A couple of updates for review @gpxl-dev. 

- Updated the typing of SubmittedTransaction to account for both approvals and orders. 
- Approval Button loading is based off of `getApprovalStatus` function in `Orders.tsx`
- Updated transactions state within `Transactions.tsx`, as I thought it would best fit where transactions would be loaded in.
- on page reload (useEffect), transactions state is preloaded from localStorage. 
- on page reload, transactions are searched to find any "processing" -> getTransactionReciept is called to find the status, and update the state accordingly.

A TODO item is figuring out how to poll every so often for any transactions. I wasn't sure how to do that, so I wanted to circle back to you. 

Let me know what you think about the changes and feedback is welcome : ).

gpxl-dev:
@codyenokida this looks really good so far.

Re: polling, if `receipt` is `null`, then I would call `getTransaction` - this will either give you a transaction from the mempool that you can call `wait` on, or `null`, in which case you need to set up an interval. Something like this (rushed, untested code!):

```typescript
transactions.forEach(async (tx) => {
  let receipt = await getTransactionReceipt(tx.hash);
  if (receipt !== null) {
    // dispatch relevant action
    return
  } else {
    // transaction not mined
    const transaction = await getTransaction(tx.hash);
    if (transaction) {
       try {
         await transaction.wait(1);
         // dispatch mine
       } catch (e) {
         // dispatch revert
       }
      return
    } else {
      // assume failed after 30 mins
      const assumedFailureTime = Date.now() + 30 * 60 * 1000;
      while (receipt === null && Date.now() <= assumedFailureTime) {
          // wait 30 seconds
          await new Promise((res)=> setTimeout(res, 30000));
          receipt = await getTransactionReceipt(tx.hash);
      }
      if (!receipt || receipt.status === 0) // revert
      else // mine
    }
  }
});
```



P.S. If you have `prettier` enabled, you should enable `format on save` in vscode settings so we all get consistent code formatting. If you have that enabled already try manually forcing a format with Cmd/ctrl + shift + P and you may be prompted to select your preferred formatter (choose prettier).

src/features/transactions/Transactions.tsx
```diff
@@ -1,13 +1,47 @@
...
+    if (chainId) {
+      const transactionsLocalStorage = JSON.parse(localStorage.getItem(getTransactionsLocalStorageKey(chainId!))!);
+
+      dispatch(setTransaction(transactionsLocalStorage))
+
+      // check from all responses if one is pending... if pending, call getTransaction
+      // to see if it was a success/failure/pending. update accordingly. if pending: wait()
+      // and poll at a sensible interval.
+      // TODO: Poll at sensible intervals
+      transactions.forEach(tx => {
```
gpxl-dev, line 29: On second look at this, I don't think you should be using `transactions` here - instead you want to use `transactionsLocalStorage`. I have a feeling `transactions` will actually not be updated at this point anyway (`transactions` would be computed on line 14 when rendering, and will not have updated between 23 and 29), but the hint is that we actually don't want this `useEffect` to have a dependency on `transactions`, which it should if it uses transactions. In other words, this effect doesn't care about the redux state of transactions changing, it only cares about transactions that were pending in localstorage at the time it ran.

Two final notes:

- Our local storage keys need to be updated to have the account in them as well as the chainId, so if I switch account I only see transactions (and activeTokens) for the account I've switched to.
- When we switch chains or accounts, we really need to cancel listeners and polling for all transactions. This effect will run to set all of those back up again if we switch back. <- this one is complex, maybe leave it 'til last and we can run through this together.

codyenokida:
Some comments on the code, regarding some decisions I went down:

1. The reason why I added an additional function inside the `useEffect` was to bypass the warning of needing `transaction` as a dependency in the dependency array. We don't want the transaction to retrigger every time the `transaction` state has changed.

src/features/metadata/metadataSubscriber.ts
```diff
@@ -1,16 +1,32 @@
...
     [chainId: number]: string[];
   } = {};
+  const transactionCache: {
+    [chainId: number]: SubmittedTransaction[];
+  } = {}
   store.subscribe(() => {
-    const { wallet, metadata } = store.getState();
+    const { wallet, metadata, transactions } = store.getState();
     if (!wallet.connected) return;
+    // persist all transactions to localStorage
```
gpxl-dev, line 18: Discussion point for later - we should probably have a limit on the number of transactions we store (even if it's quite high!)
There should also be the ability to clear this. 

I think we should address both of these as a separate PR though, because otherwise this one will become a monster!

gpxl-dev (approved):
This looks awesome to me 🔥 

When I submit a transaction & refresh the page, I see the pending transaction after I reconnect my wallet and it correctly updates when the transaction mines.

It's a bit hard to test some of the edge and failure cases at the moment, but as you suggested we can add some tests in Dappetteer once we get to that point.

I made a few small changes: 

- some pedantic renames (note that the rename commit also changed formatting because I use prettier to format - you should too!)
- I added a check for `library` when you're checking `chainId` and `account`. AFAIK all 3 will always be truthy together and adding this check meant you could remove some `!` and `?` further down without typescript complaining.
- I added a boolean flag to set when the wallet changes that will prevent a subsequent dispatch happening that would only be relevant for the previous wallet's state.
- I removed your extra function - just changing from `transactions` to `localStorageTransactions` was enough.

If you have any questions about any of the above, give me a shout. Also please do test my changes!

P.S. If you'd prefer I don't commit on your branch let me know (DM me on discord!) - I just thought it would be easier for these v small changes.

src/features/orders/Orders.tsx
```diff
@@ -10,6 +10,8 @@ import {
   selectOrder,
   selectOrdersStatus,
 } from "./ordersSlice";
+import { selectTransactions } from "../transactions/transactionsSlice";
```
hmalik88, line 13: This can be a single line

src/features/transactions/Transactions.tsx
```diff
@@ -1,13 +1,107 @@
...
+          getTransactionsLocalStorageKey(chainId!, account!)
+        )!
+      ) || { all: [] };
+
+      dispatch(setTransactions(transactionsLocalStorage));
+
+      // check from all responses if one is pending... if pending, call getTransaction
+      // to see if it was a success/failure/pending. update accordingly. if pending: wait()
+      // and poll at a sensible interval.
+      transactionsLocalStorage.all.forEach(async (tx) => {
```
hmalik88, line 36: It might be worth it to abstract away the callback to its own named function (not sure if in the same file or elsewhere) to improve readability. Maybe something like `processTransactions`.

PR #62, opened 2021-06-24
Quote timer

codyenokida:
Quote Timer is a component that takes in a Unix timestamp (acquired from order request) and counts down to 0:00.

Functionalities include:
1. counting down
2. calling request() when the timer hits 0
3. stopping when order is taken (further behavior might change when designs get finalized -> redirected to a new page state)


gpxl-dev (changes requested):
1. Prop name of `unixTimestamp` is not perfect - having looked at the code I can now see this is the expiry time, but I think you could name this better to make it clearer. Check `button.tsx` to see how you can write a description on props that you think may require some additional explanation (e.g. if you called it `expiryTime` you may wish to add in the description that it expects a unix timestamp). Similarly I'm not sure if you need the `timer` part of `timerDisabled` but my opinion about that is less strong!

2. I think within your interval it would be better to calculate the `distance` every time, then work out seconds and minutes from there, rather than calculate minutes and seconds once and decrement. This is because if there is any error in timing, it will be compounded and never rectified if you use the relative approach. As a point of note (you are probably already aware!), Javascript intervals and timeouts are not guaranteed to run after exactly the time specified, e.g. if you set a timeout for 1000ms from now, it can run any time _after_ 1000ms from now so for a longer timer it _might_ add up to be significant. It also makes things easier in the case of your disable button. Currently there's a weird bug where if you disable then re-enable the timer it starts again from the top of the current minute (try it in storybook). I think this would fix itself if you just used the expiry. You may want to add a test for disabling/re-enabling if you think it is important functionality.

3. Don't check `=== 0`, check `<=0` - sometimes you can get into a situation where the timer ticks past zero (I think I disabled and re-enabled after 5 mins later). This results in some weird behaviour: 
image

4. Would be good to understand the use-case for `disabled` (I expect we'd probably just conditionally remove the timer), but I'd argue that showing `0:00` is misleading. Not a strong opinion.

src/components/Timer/Timer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  timerDisabled,
+}: TimerProps) => {
+  const [minutes, setMinutes] = useState<number>(0);
+  const [seconds, setSeconds] =  useState<number>(0);
+
+  
+  useEffect(() => {
+    const now = new Date();
+    const unixDate: Date = new Date(unixTimestamp * 1000);
+    const distance = (unixDate.getTime() - now.getTime()) / 1000;
```
hmalik88, line 22: why do you have `unixTimestamp * 1000` on line 21, shouldn't the unixTimestamp already be in milliseconds? I'm not exactly sure if the expiry we get back in our orders is in seconds/ms. Regardless, if the component doesn't receive a unixTimestamp param, the timestamp is in ms and the useEffect is still multiplying by 1000.

hmalik88 (commented):
Agree w/ stuff Ed said, made a comment about a line in `Timer.tsx`

src/components/Timer/Timer.tsx
codyenokida: Yep, unixTimestamp is in s, not ms. I'll reflect this change in the next commit :)

hmalik88 (approved):
LGTM

gpxl-dev (approved):
🔥

hmalik88 (approved):
👍

PR #63, opened 2021-06-26
Fixed bug; transactions accidentally persisted in localStorage throug…

codyenokida:
Oops, persisting transactions to localStorage had a minor bug!

To reproduce the bug:
1. make any transaction
2. switch chains (Rinkeby -> Mainnet)
3. transactions persist and localStorage for Mainnet is overridden by the one on Rinkeby
4. refreshing the page will still persist transactions that should've happened on Rinkeby on the Mainnet ui

To overcome the bug, I added a check to listen to transaction changes and updated the transactionCache structure. We might get a minor performance increase if we add a check to handle transaction changes, compared to the previous commit.

### DIFF
--- a/src/features/metadata/metadataApi.ts
+++ b/src/features/metadata/metadataApi.ts
@@ -23,9 +23,11 @@ const tokensCache: {
   [chainId: number]: Promise<TokenInfo[]>;
 } = {};
 
-export const getActiveTokensLocalStorageKey: (chainId: number) => string = (
-  chainId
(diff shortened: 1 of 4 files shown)


gpxl-dev (approved):
Nice find - can confirm the fix works.
I also hijacked this PR to update the `activeTokens` metadata listener to store against account as well as chain.

hmalik88 (approved):
LGTM 🤙

PR #64, opened 2021-06-27
Issue 55: Add visual indicator if sender amount is greater than current balance

codyenokida:
New pr 👯‍♂️ 

Added feature to enable red text whenever senderAmount is greater than the current balance. 

Features:
1. Input text turns red if senderAmount is greater than the current balance.
2. Request button texts changes to have 3 options: "Enter an Amount" (initial), "Request" (sufficient), "Insufficient ___ balance" (insufficient)
3. Added `handleTokenAmountChange` function to handle non numeric and period values into input.
4. new AmountStatus type in `TokenSelect.tsx`

Some discussion points:
1. Do all tokens have 18 decimal denominations? I assumed they do, so calculate balance and current senderAmount based off of that.
2. Should a user still be able to request a quote if they don't have enough balance?
3. Is AmountStatus a good name for the status of the senderAmount? (I'm thinking we should change it to something better but idk what)

Please leave some feedback!

### DIFF
--- a/public/locales/en/orders.json
+++ b/public/locales/en/orders.json
@@ -7,5 +7,7 @@
   "send": "Send",
   "receive": "Receive",
   "amount": "Amount",
-  "chooseToken": "Choose token"
+  "chooseToken": "Choose token",
+  "enterAmount": "Enter an amount",
+  "insufficentBalance": "Insufficient {{symbol}} balance"
 }
--- a/src/components/TokenSelect/TokenSelect.tsx
+++ b/src/components/TokenSelect/TokenSelect.tsx
@@ -4,8 +4,6 @@ import { HiOutlineChevronRight } from "react-icons/hi";
 import { useTranslation } from "react-i18next";
 import LoadingSpinner from "../LoadingSpinner/LoadingSpinner";
 
-export type AmountStatus = "initial" | "sufficient" | "insufficient";
-
 type TokenSelectPropTypes = {
   withAmount: boolean;
   label: string;
@@ -13,15 +11,9 @@ type TokenSelectPropTypes = {
   amount?: string;
   token?: string;
   tokens: TokenInfo[];
-  amountStatus?: AmountStatus;
+  hasError?: boolean;
   onAmountChange?: React.FormEventHandler<HTMLInputElement>;
   onTokenChange?: React.FormEventHandler<HTMLSelectElement>;
-};
-
-const colorClasses: Record<AmountStatus, string> = {
-  initial: "dark:text-white",
-  sufficient: "dark:text-white",
-  insufficient: "dark:text-red-700 text-red-700",
 };
 
 const TokenSelect = ({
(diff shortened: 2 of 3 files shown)


gpxl-dev:
To your questions:

> 1. Do all tokens have 18 decimal denominations? I assumed they do, so calculate balance and current senderAmount based off of that.

No, e.g. `AST` has `4`. I think 18 is the most common though.

> 2. Should a user still be able to request a quote if they don't have enough balance?

My instinct is to agree with what you have done (i.e. no), but we can raise this to a larger audience.

> 3. Is AmountStatus a good name for the status of the senderAmount? (I'm thinking we should change it to something better but idk what)

Personally I would keep this generic and just go for something like `error` or `hasError` and make it a boolean. If there are other non error stats (e.g. warn), we could add others later.

gpxl-dev:
In terms of implementation, I am not sure if we need a `useEffect` for this? Could we just have something like:

```javascript
const parsedSenderAmount = parseUnits(senderAmount, senderTokenInfo.decimals);
const insufficientBalance = BigNumber.from(balances?.values[senderToken]).lt(parsedSenderAmount);

// later

<TokenInput hasError={ /* some things and */ insufficientBalance} />
```

It would run on every render but I think everything there is super quick to deal with.

P.s. it's `parseUnits` you need, and the `decimals` comes from the `TokenInfo` which you should be able to get from `metaDataSlice`

codyenokida:
> const parsedSenderAmount = parseUnits(senderAmount, senderTokenInfo.decimals);
Where did you get this `parseUnits` from? Would toAtomicString not work in this case if it is not 18 decimal places?

> Personally I would keep this generic and just go for something like error or hasError and make it a boolean. If there are other non error stats (e.g. warn), we could add others later.

Would an enum type like what I added in the initial commit work for non-error stats? Or would you guys prefer a different method in doing that? I wanted to display "Set Enter Amount" when the senderAmount was 0, but we can toggle that text without having the enum type (e.g. "initial"), so would want to clarify this! 

> In terms of implementation, I am not sure if we need a useEffect for this? Could we just have something like:
Is it necessary to run on every render? I thought useEffect would be good to listen into the dependency changes when a user switches amounts/tokens. If your implementation is best practice then I'd go for that one :D

gpxl-dev:
> thought useEffect would be good to listen into the dependency changes when a user switches amounts/tokens. If your implementation is best practice then I'd go for that one

 Yes - thought about this more. As per discord - `useEffect` is for side effects specifically and will actually have a negative performance impact if you use it to recompute values when things change (because it will cause an additional render - the `useEffect` callback runs after a render, then triggers a `setState` which causes another render).

Your "not enough funds" state is "computed state" - it can be calculated based exclusively on other known variables. The general rule for computed state is if it's 'cheap', just compute it as a part of the render and if not, memoize it (`useMemo`). 

> I wanted to display "Set Enter Amount" when the senderAmount was 0, but we can toggle that text without having the enum type (e.g. "initial"), so would want to clarify this!

Sorry - I should have been clearer. I was only talking about `TokenSelect`s props - I don't think it needs to differentiate between the different states you have in `Orders.tsx` because it is only changing the text colour. For the "initial" state, I would expect an input to have either a `pristine` or `dirty` prop (with dirty being the opposite of pristine - i.e. untouched vs touched).

In terms of the text on the button, what you have works fine, so stick with it if you like it the most (although get rid of the `useState`, as again this can be computed (`!senderAmount || senderAmount === "0"`)

PR #73, opened 2021-07-08
Feature: Token Selection Modal

codyenokida:
Hi there :)

Token selection modal takes in active tokens and balances to create a modal screen to select sender and signer tokens. Some notable features:
1. searchable list based off of symbol and name text
2. searchable list based off of contract id
3. list sorted based off of token balance first, then alphabetically by the name of symbol
4. sender and signer token swaps when counter part token is already selected; (e.g. senderToken (DAI) and signerToken (WETH) swaps values if user selects signerToken as DAI. *end result: senderToken (WETH), signerToken (DAI)*)
5. selected prop in token row: (if counterpart token is already selected) (e.g. signerToken === DAI, DAI is "selected" when selecting senderToken)
6. disabled prop in token row: (if token is already selected, disable that selection) (e.g. senderToken === DAI, DAI is disabled when selecting senderToken)

Some TODOS (that I think should be linked in a separate PR):
1. styling (using styled-components) or whatever styling choice we use
2. adding maker supported tokens
3. adding ability to add inactive tokens

**Currently set as a modal, but subject to change when designs are finished. **


gpxl-dev:
Hi 👋 ,

1. I think `sort.ts` and `filter.ts` are good candidates for unit tests.
2. Sorting:
    1. I'm not sure if it works as you intended:
    image
    2. As I think Don mentioned in our call - it's probably best not to sort by number of tokens - prefer alphabetical for now, with a longer term goal to potentially sort by USD value. Compare holding 10ETH with 11AST - which should be at the top of the list?
3. We should incorporate the ability to add to active tokens in this modal - question around how exactly this should work but perhaps initially if an address is pasted into the filter box that matches another known token (from metaDataSlice) it gives the option to select and/or add to token list.
4. Would be nice to have a story for this, but you will have to read up on how to get redux state into storybook (or how to mock it).

gpxl-dev (changes requested):
see comment

src/components/TokenSelect/TokenSelect.tsx
```diff
@@ -85,21 +84,10 @@ const TokenSelect = ({
...
-              </option>
-              {tokens.map((token) => (
-                <option key={token.address} value={token.address}>
-                  {token.symbol}
-                </option>
-              ))}
-            </select>
+              onClick={onTokenChange}
+              >
+              {tokens.find((t)=> t.address === token)?.symbol || isDefaultOptionDisabled ? tokens.find((t)=> t.address === token)?.symbol : `…${t("common:select")}`}
```
hmalik88, line 89: You can move the ellipsis to after the select, it shows as "...Select" when it should be "Select...", I think the way we had it before was necessary for the option select tag, I guess for button tags it should be done normally.

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -31,6 +31,12 @@ import Timer from "../../components/Timer/Timer";
 import Modal from "react-modal";
 import Card from "../Card/Card";
 import WalletProviderList from "../WalletProviderList/WalletProviderList";
+import TokenSelection from "../../components/TokenSelection/TokenSelection";
+
+export interface TokenSelectionModalState {
```
hmalik88, line 36: Where are you using this interface?

src/features/orders/Orders.tsx
```diff
@@ -1,5 +1,9 @@
 import SwapWidget from "../../components/SwapWidget/SwapWidget";
 
+export interface TokenSelectionState {
```
hmalik88, line 3: Same for this one, where are you using this?

hmalik88 (changes requested):
LGTM apart from the comments Ed made and the minor comments I left.

src/components/SwapWidget/SwapWidget.tsx
codyenokida: Had an idea to use an interface - but scratched it and forgot lol. thanks for pointing it out :D

codyenokida:
Final additional changes; added ability to query through inactive token lists to add to the active tokens. 

Default active tokens will not be able to be removed from active token list, but all other tokens will have the option to do so. 

Storybook support and added a couple different states each component could be in :)

hmalik88 (changes requested):
Hey! Found some errors and areas where we can potentially improve the UX:

**Sugggestions**

1. @dmosites, not sure if you wanted the token selection to be a drawer in this PR. If so, this component should probably be taken out of the modal and some transition between the states should be added?
2. When you remove a token from the active token set, the modal closes, it should remain open for visual confirmation that it's removed and for the user to have a chance to choose a token.
3. IMO, we should not disable the option to remove a token from active tokens if it’s selected, if a person does remove a token from active tokens and it’s currently selected, the state should be reset and the modal should remain open. I think there should also be some visual cue that lets a person know they're clicking "remove token" and not actually choosing the token (I guess an underline on hover/click would suffice?).
4. Expanded results from inactive token lists text still remains, not sure if we want to keep that text or have some other visual separation between the two lists.
5. Not sure if you're using `prettier` on save, but I think I saw some LOC that looked a bit long.

**Errors**

6. When you remove a token from active tokens the state is reset and the `SwapWidget` button hits a ‘Token decimals not found’ state, even while another token is selected.
7. The `TokenRow`s  and `TokenImportRow`s should have a key prop in `TokenSelection`.
8. I think you left some stray console.logs in `TokenSelection`.
9. Example of other erroneous behavior I observed, which is tied to point 6 ^: If I choose AST for sender, then for signer I search and add APY to active tokens, choose APY for sender, and finally choose to remove APY from the sender token select, the `SwapWidget` button goes to the "Token decimals not found" state, the sender becomes unselected and signer now becomes AST.

codyenokida:
@hmalik88 thanks for the feedback! 

I fixed the changes and errors that you mentioned. I realized that the onClick event on the removeActiveTokens text was firing the parent-on-click event. now that's fixed, so it fixes up the UX issues that you brought up!

src/components/TokenSelection/TokenSelection.tsx
```diff
@@ -0,0 +1,194 @@
...
+  setSenderToken,
+  tokenSelectType,
+  balances,
+  allTokens,
+  activeTokens = [],
+  addActiveToken,
+  removeActiveToken,
+  chainId,
+}: TokenSelectionProps) => {
+  const [tokenQuery, setTokenQuery] = useState<string>("");
```
piersss, line 80: `useState("")`

Defines the type as string too.

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -31,23 +40,31 @@ import Timer from "../../components/Timer/Timer";
...
+import TokenSelection from "../../components/TokenSelection/TokenSelection";
 
 const floatRegExp = new RegExp("^([0-9])*[.,]?([0-9])*$");
 
 const SwapWidget = () => {
   const [senderToken, setSenderToken] = useState<string>();
   const [signerToken, setSignerToken] = useState<string>();
-  const [senderAmount, setSenderAmount] = useState("0.01");
+  const [senderAmount, setSenderAmount] = useState<string>("0.01");
   const [showWalletList, setShowWalletList] = useState<boolean>(false);
```
piersss, line 51: `useState(false)`

Defines the type as boolean too.

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -31,23 +40,31 @@ import Timer from "../../components/Timer/Timer";
...
 import WalletProviderList from "../WalletProviderList/WalletProviderList";
+import TokenSelection from "../../components/TokenSelection/TokenSelection";
 
 const floatRegExp = new RegExp("^([0-9])*[.,]?([0-9])*$");
 
 const SwapWidget = () => {
   const [senderToken, setSenderToken] = useState<string>();
   const [signerToken, setSignerToken] = useState<string>();
-  const [senderAmount, setSenderAmount] = useState("0.01");
+  const [senderAmount, setSenderAmount] = useState<string>("0.01");
```
piersss, line 50: Use number type for senderAmount

src/components/TokenSelection/TokenImportRow.tsx
```diff
@@ -0,0 +1,64 @@
...
+        src={token.logoURI || "https://via.placeholder.com/150"}
+        className="w-6"
+        alt="hello"
+      />
+      <div className="flex flex-col justify-start">
+        <h3 className="flex flex-col">{token.symbol}</h3>
+        <h3 className="text-gray-400">{token.name}</h3>
+      </div>
+      <span></span>
+      <button
```
piersss, line 54: `<button disabled />`

Give button a disabled prop and there is no need to check the disabled component prop.

src/components/TokenSelection/TokenRow.tsx
```diff
@@ -0,0 +1,90 @@
...
+  token,
+  balance,
+  setToken,
+  removeActiveToken,
+  selected,
+  disabled,
+  defaultToken,
+}: TokenRowProps) => {
+  return (
+    <div
```
piersss, line 45: This should be button, not a div. Then same as above, add disabled prop to button

src/components/TokenSelection/TokenSelection.tsx
```diff
@@ -0,0 +1,194 @@
...
+  return (
+    <div>
+      <div className="flex flex-wrap align-middle justify-between">
+        <h1 className="font-bold text-sm">Select token</h1>
+        <HiX
+          className="light:text-black text-xl cursor-pointer"
+          onClick={closeModal}
+        />
+      </div>
+      <input
```
piersss, line 131: This input should have a label.

src/components/TokenSelection/TokenSelection.tsx
```diff
@@ -0,0 +1,194 @@
...
+        type="text"
+        value={tokenQuery}
+        placeholder="Search name or paste address"
+        onChange={(e) => {
+          setTokenQuery(e.target.value);
+        }}
+        className="w-full"
+      />
+      <div>
+        {filteredTokens.map((token) => {
```
piersss, line 141: `{filteredTokens.map((token) => ((
    <TokenRow`

piersss (changes requested):
Hey there Cody. Looks good! Some other small things not part of this feature, but hopefully you can fix this too:

1. TokenSelect onAmountChange should return a number, not an event
2. Same for onTokenChange, it should return TokenInfo, not an event
3. TokenSelect input should have `pattern=“^[0-9]*[.,]?[0-9]*$”`, in this way you don’t have to filter any characters in javascript. Uniswap app's input is like this: 

`<input class="sc-1x3stf0-0 kjVGCt token-amount-input" inputmode="decimal" autocomplete="off" autocorrect="off" type="text" pattern="^[0-9]*[.,]?[0-9]*$" placeholder="0.0" minlength="1" maxlength="79" spellcheck="false" value="4.989541028995924952">`

src/components/TokenSelection/TokenImportRow.tsx
codyenokida: One point for discussion here: TokenImportRow has no onclick events when clicking on the row itself, so should it even be wrapped as a button? For uniformity sake, I guess it should.

Also there shouldn't actually be a need for disabled or selected prop for importing tokenImportRow, since the only purpose it serves is to import the token.

src/components/SwapWidget/SwapWidget.tsx
hmalik88: I believe we've kept this as string because `getSignerSideOrder` is expecting a string

src/components/TokenSelection/TokenImportRow.tsx
piersss: Ah yes I see now. I didn't notice that there is no onClick event listener. Also according to W3C recommendation you shouldn't wrap a button in a button. So in this case you should keep it as div. My bad!

src/components/SwapWidget/SwapWidget.tsx
piersss: Ah ok then it can stay like this.

src/components/TokenSelection/TokenRow.tsx
piersss: I noticed that there's a <span> (removeActiveToken) inside the component that functions like a button. So we have a button in a button right now which is not recommended. You have to wrap the component in a container and add the two buttons on the same level.

piersss: If this is too much work then I could also do it when I do the styled-components rewrite.

codyenokida: Would you be able to do it, or guide me to how this should be implemented? We can perhaps do it in a separate issue as well too

piersss: Alright, let's keep it like this for now.

src/components/TokenSelection/TokenImportRow.tsx
```diff
@@ -0,0 +1,64 @@
...
+   */
+  token: TokenInfo;
+  /**
+   * onClick event, either setSignerToken or setSenderToken
+   */
+  onClick: (val: string) => void;
+  /**
+   * Whether counterpart token is already selected; Still allows user to select that token
+   */
+  selected?: boolean;
```
redders6600, line 17: As per comment on TokenRow

src/components/TokenSelection/filter.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    const filteredTokens = filterTokens(activeTokens, "0x");
+    expect(filteredTokens.length).toBe(1);
+    expect(filteredTokens[0].symbol).toBe("ZRX");
+  });
+
+  it("should filter tokens by token address (0x2260fac5e5542a773aa44fbcfedf7c193bc2c599)", () => {
+    const filteredTokens = filterTokens(activeTokens, "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599");
+    expect(filteredTokens.length).toBe(1);
+    expect(filteredTokens[0].symbol).toBe("WBTC");
+  });
```
redders6600, line 45: Fine for now - could add a couple more test cases in here though in future, e.g. enter:

- nothing
- one character
- something that matches the middle of a symbol (e.g. `ET` in `WETH`)

redders6600 (approved):
Nice! Working well functionally, approving but could benefit from a name change on your token row components IMO (see comment).

Will also commit a couple of minor typing changes.

src/components/TokenSelection/TokenRow.tsx
```diff
@@ -0,0 +1,91 @@
...
+   */
+  balance: string;
+  /**
+   * onClick event, either setSignerToken or setSenderToken
+   */
+  setToken: (val: string) => void;
+  /**
+   * Whether counterpart token is already selected; Still allows user to select that token
+   */
+  selected?: boolean;
```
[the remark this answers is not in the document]
codyenokida: I looked into what uniswap does, and they take a similar approach is restyling the "other" token. But I agree, I don't think it's super necessary; I deleted the prop for now and if in the designs we have the need to have another state for when the "other" token is selected, we can add it back with a better name hehe

piersss (approved):

hmalik88 (approved):

PR #75, opened 2021-07-12
Feature/styled components

piersss:


piersss:
I also added dark / light mode, but not to the store yet. Should we add userSettings or something to the store for this?

codyenokida:
Hey! Apologies for the late review. 

Styled components integration looks great! Great start to the user interface. I think it's almost ready to push, but I think we should try and squash a couple more bugs before we are ready to push to dev. We probably would not need to kill all these issues in this PR. The smaller details or updates along the way could be split up into separate issues/tickets as we move on throughout the dev process. I know you are still waiting on the designs, so sorry if some of the bugs/errors are out of your control! Just want to keep a note of them.

**A note of some bugs I have seen:**
1. Mobile view of swap widget and logo placement there
2. Clicking on the sidebar items shouldn't cause a refresh of the page
3. Follow up to 2: clicking the sidebar items resets the dark/light theme (this should be stored in local or session storage)
4. styled-components not working in storybook light/dark mode.

**A couple of comments and discussion points for the other devs to contribute.**
1. Should the wallet connect button be inside the sidebar or outside?

**Questions I have (just cus I'm curious and you probably have ore experience in styled-components more than I do)**
1. Not a big deal, but in `TraderContainer.tsx` should we avoid using className="container-children"?  I was under the impression that when using styled-components, it's good practice to make all your components styled-components. (e.g. separate .container-children into it's own styled.div.)

piersss:
Thanks for the feedback cody.

1. Mobile has no design yet, so still waiting on that. I think restyling the swap widget should be in another feature.
2. Sidebar is also temporary, but will fix the routing issues.
3. Oh yeah, forgot to write to localStorage when switching theme!
4. Will look into that.

I've added userSettings to store, theme is saved in there.

Only the parent of a component should be a StyledComponent. If every child is a StyledComponent then you have to pass all props regarding styling individually to each child. This will get messy really fast. Styling by child classname makes things a lot easier.

redders6600:
We should definitely use [the babel macro](https://styled-components.com/docs/tooling#babel-macro) for all of the reasons mentioned at the top of that page. It should "just work" with CRA by changing all of the imports from `styled-components` to `styled-components/macro`.

redders6600:
>If every child is a StyledComponent then you have to pass all props regarding styling individually to each child.

You can interpolate a child component in the parent's styles, I think, like this:

```
const Child = styled.span`
  font-weight: bold;
`

const Parent = styled.div`
  ${Child} {
    text-transform: uppercase;
  }

`

**Comment by piersss** [comment]
> > If every child is a StyledComponent then you have to pass all props regarding styling individually to each child.
> 
> You can interpolate a child component in the parent's styles, I think, like this:
> 
> ```
> const Child = styled.span`
>   font-weight: bold;
> `
> 
> const Parent = styled.div`
>   ${Child} {
>     text-transform: uppercase;
>   }
> 
> `
> ```

I could implement this, but converting all childs to components won't make it immediately obvious which elements are simple div's and span's, and which are components.

```
<StyledComponent>
  <div className="text">
    Text
  </div>
  <ProgressBar />
  <div className="image-container">
      <img className="image" src="image.jpg" />
  </div>
</StyledComponent>
```

Compare to:

```
<StyledComponent>
  <Text>
    Text
  </Text>
  <ProgressBar />
  <ImageContainer>
    <Img src="image.jpg" />
  <ImageContainer/>
</StyledComponent/

```

It looks cleaner but as you can see it's not immediately obvious which elements are components and which are simpele html elements.

PR #86, opened 2021-07-23
Feature/swap widget styling

piersss:
Restyled SwapWidget with styled-components
Added TextInput component for text and number inputs
Added global styling for react-modal
Added alwaysWhite and alwaysBlack to theme colors


src/components/TokenSelect/TokenSelect.styles.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  ${StyledTextInput} {
+
+    label {
+      margin-bottom: 0.125rem;
+    }
+
+    input {
+      padding-bottom: 0.5rem;
+    }
```
piersss, line 57: Is this the way to style the children of another component?

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { FC, ReactElement } from 'react';
+import { StyledFormLabel, StyledInput, StyledTextInput } from './TextInput.styles';
+
+type HTMLInputProps = Omit<React.HTMLProps<HTMLInputElement>, 'ref' | 'as'>;
```
piersss, line 4: Could not find the right types for the input element returned by styled-components. Any way to improve this?

gpxl-dev: Can you define the type as the following:

```typescript
export type TextInputProps = {
  label: string;
  type?: string;
  hasError?: boolean;
  hideLabel?: boolean;
} & typeof StyledTextInput;
```

src/components/TokenSelect/TokenSelect.styles.tsx
gpxl-dev: To me it feels weird to style the child element of another component. I think it's DOM structure should be irrelevant to other components.

I think there are three options:

1. Preset layouts for the component as above
2. Expose several classNames style props (e.g. labelClassNames, inputClassNames) which accept a css class name and apply it to the label. Not 100% sure how this works with styled components but assume something to that effect is possible.
3. (Similar to 2), expose a labelCss prop (and one for the input) which expects styled component css from the css utility (FlattenSimpleInterpolation seems to be the type). Ref here: https://github.com/airswap/airswap-web/blob/e9e6750dcec2d976a55638e5a1e806f100ea016a/src/components/Typography/Typography.styles.tsx#L59

src/components/TextInput/TextInput.tsx
piersss: I tried this but it won't give me the right input element props. My solution works but I find it strange I have to omit certain values.

src/components/TokenSelect/TokenSelect.styles.tsx
piersss: I found a solution, in TokenSelect.styles.tsx you can import the TextInput child elements:

```
export const StyledTokenSelect = styled.div<StyledTokenSelectProps>`

  ${StyledTextInput} {
    
    ${StyledFormLabel} {
      margin-bottom: 0.125rem;
    }
    
    ${StyledInput} {
      padding-bottom: 0.5rem;
    }    
  }
}

```

gpxl-dev (changes requested):
All looks good except for the "Receive" heading - it has 0.5 opacity, therefore looking grey doesn't match the designs.

This is because the text input is set to be disabled. In this particular case, the label applies to both a number input (which doesn't really exist on receive) and a select button. Given the select button is active, I don't think we should add disabled styling to the label so I agree with the design

codyenokida (changes requested):
LGTM  👍

codyenokida (approved):
LGTM 👍

PR #117, opened 2021-08-12
Feature/token selector sorting and filtering

piersss:
- Added performance improvement for TokenSelection, loading time went from 40.5 ms to 12.1 ms.
- Added sorting function for tokens: non-null balanced tokens sort on top.

### DIFF
--- a/src/components/TokenSelection/sort.ts
+++ b/src/components/TokenSelection/sort.ts
@@ -27,6 +27,12 @@ export function sortTokensBySymbol(tokens: TokenInfo[]) {
 }
 
 function getTokenBalance(token: TokenInfo, balances: BalancesState): number {
+  const balance = balances.values[token.address];
+
+  if (!balance) {
+    return 0;
+  }
+
   return parseFloat(
     formatUnits(balances.values[token.address]!, token.decimals)
   );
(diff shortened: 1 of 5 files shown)


codyenokida:
I get an error when importing new tokens. Perhaps you need to add error checking in 

` // sort tokens based on symbol
  const sortedTokens: TokenInfo[] = useMemo(() => {
    return sortTokensBySymbolAndBalance(activeTokens, balances);
  }, [activeTokens, balances]);`

`Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.1.1)`

src/components/TokenSelection/sort.ts
```diff
@@ -25,3 +25,27 @@ export function sortTokensBySymbol(tokens: TokenInfo[]) {
     a.symbol.toLocaleLowerCase() < b.symbol.toLocaleLowerCase() ? -1 : 1
   );
 }
+
+function getTokenBalance(token: TokenInfo, balances: BalancesState): number {
+  return parseFloat(
+    formatUnits(balances.values[token.address]!, token.decimals)
+  );
+}
+
```
codyenokida, line 34: This is where the undefined BigNumber error gets triggered, causing an app crash

codyenokida (changes requested):

src/components/TokenSelection/sort.ts
piersss: Fixed, nice find.

PR #132, opened 2021-08-18
[Alpha] Content side bar 

codyenokida:
Hi! This is a PR to add the new content side bar. 

Much of this is probably still subject to change (such as background color, animation style, etc.), but interested to get some feedback to see if this is ready to be merged in.

Features:
1. Ability to retract side bar
2. minor routing (swap/introduction/products/organizations)
3. animation using framer-motion (thanks ed for inspiration from your branch!)
4. minor style changes of the entire page, so swap widget will be centered

TODOS: mobile view for this (currently display: none)


dmosites:
Looking good, would just like to remove the bounce for now. Also noticing the :focus draws a square around the button, maybe we drop focus styling until we do an accessibility pass?

src/components/Page/Page.styles.tsx
```diff
@@ -10,17 +10,22 @@ export const StyledSiteLogo = styled(SiteLogo)`
...
+export const StyledPage = styled.div<StyledPageProps>`
   display: flex;
   justify-content: center;
-  align-items: flex-start;
+  align-items: center;
   position: relative;
   min-width: 18rem;
   min-height: 100vh;
   padding: 2rem 0;
+  transition: padding-right 0.3s ease-in-out;
```
gpxl-dev, line 25: This not a performant way to animate. Ideally you want to be animating (transitioning) either `transform` or `opacity`. There's some info [here](https://web.dev/animations-guide/).

Same applies to the wallet positioning (which I think has `transition: margin-right`).
On my currently overworked macbook, the wallet and swap widget aren't particularly smooth.

It's definitely fine for the alpha, but something we should address before public beta (which will also have a mobile layout, so should help with some of the layout issues you might face if you switched to `transition: transform` now).

gpxl-dev (approved):
LGTM with a note about animation performance - agree with Don's comments though!

PR #135, opened 2021-08-19
Restrict tokens to those in registry

### LINKED ISSUE #126: Add hard coded list of supported tokens and disable other tokens on the token selector

gpxl-dev:
- Adds a `registry` slice to state which contains a list of the tokens supported by each staker (maker), as well as a list of all tokens supported across all makers.
- Updates the token selection screen to prevent importing unsupported tokens to a user's active tokens list
- Upgrades @airswap/protocols

This branch probably has a non-ideal UX, but preventing active token import is the fastest & most feasible way to implement pre-alpha.

Screenshot:
image

Closes #126

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@airswap/balances": "^1.1.2",
     "@airswap/constants": "^0.4.0",
     "@airswap/metadata": "^0.3.2",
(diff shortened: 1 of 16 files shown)


dmosites (approved):
LGTM

PR #138, opened 2021-08-19
Don't show submitted screen if signature rejected

gpxl-dev:
Previously rejecting a metamask signature request would essentially be ignored and incorrectly show the transaction submitted screen.

Note for anyone reading this who is new to redux toolkit - I had wrongly assumed that `await dispatch(someThunkGeneratingAction)` would throw if the thunk rejected. It doesn't - you need to use `unwrapResult`, like this:

```javascript
try {
  const result = await dispatch(someThunkGeneratingAction)
  await unwrapResult(result);
  // do success specific stuff.
} catch (e) {
  // you would never get here if you didn't await the unwrap.
}
```

### DIFF
--- a/src/components/SwapWidget/SwapWidget.tsx
+++ b/src/components/SwapWidget/SwapWidget.tsx
@@ -202,8 +202,17 @@ const SwapWidget = () => {
             disabled={isNaN(parseFloat(signerAmount))}
             loading={ordersStatus === "taking"}
             onClick={async () => {
-              await dispatch(take({ order, library }));
-              setShowOrderSubmitted(true);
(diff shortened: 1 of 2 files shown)


gpxl-dev:
P.S. I'm not sure if there are other reasons submitting the transaction will fail (I know it happens if the arguments are wrong according to the ABI, but I don't think this should occur with our code), but if there are we probably ought to notify the user. Strictly speaking I'm not sure this falls into our agreed toast categories, but for now perhaps we can make it a toast (once they're merged).

dmosites (approved):
LGTM

PR #141, opened 2021-08-19
feature/swap-widget-rounded-corners

piersss:
Added border-radius to TradeContainer


piersss:
Added border-radius to TradeContainer

PR #148, opened 2021-08-20
[Feature] - Toast Notification

codyenokida:
Fixed up the styling for toast notification! 

- changed styling from Jop's designs
- removed generic toast
- edit transaction toast
- added i18n text


src/components/Toasts/Toaster.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import styled from "styled-components/macro";
+
+const ToasterWrapper = styled.div`
+  div[role="status"] {
+    margin: 0;
+  }
+  div {
+    max-width: 100%;
+    border-radius: 8px;
```
gpxl-dev, line 11: Wonder if this should be 0.5rem maybe?

src/components/Toasts/TransactionToast.styles.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from "styled-components/macro";
+
+export const Container = styled.div`
+  color: ${(props) => props.theme.colors.white};
+  background-color: ${(props) => props.theme.colors.darkGrey} !important;
```
gpxl-dev, line 5: There are a few `!important` rules in here - as I'm sure you know this is generally discouraged - are you overriding some built-in-style or something?

I think if we need to use `!important` there should be a comment to explain exactly why - I feel like it should be very rare with styled components.

src/components/Toasts/TransactionToast.styles.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from "styled-components/macro";
+
+export const Container = styled.div`
```
gpxl-dev, line 3: @piersss will _not_ be happy the ordering of the css properties in this file! :sweat_smile:

src/components/Toasts/TransactionToast.styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 2.5rem;
+  height: 2.5rem;
+  color: white;
+  color: ${(props) => (props.error ? "#FF0000" : "#60FF66")};
+`;
+
+export const HiXContainer = styled.div`
```
gpxl-dev, line 32: Semantically I think this should be a button

gpxl-dev (changes requested):
Looks great - some really minor styling changes but functionally works really well!

gpxl-dev:
After our call:
https://react-hot-toast.com/docs/styling

See "Change the position of the toaster". 
Believe Don was suggesting putting it in the application frame (i.e not overlapping the content pane when open).

PR #163, opened 2021-08-21
fix/loading-flicker

piersss:
- Added PageLoader component
- Fixed small SideBar isOpen state bg

Instead of Loading text you get this:
Screenshot 2021-08-21 at 15 33 48


gpxl-dev (approved):
This is great - as an aside I think we should still track having some `<Suspense>` components lower down the tree to catch places that i18n translations are lazy loaded in after the initial page load.

PR #178, opened 2021-08-26
Added token sorting based on symbol first, then substring

codyenokida:
Checked out other protocols to see a good optimized way to do this :) 

- filter based on symbol first

Some other unrelated issues:
- Uniswap token needs a placeholder

### DIFF
--- a/src/components/TokenSelection/TokenSelection.tsx
+++ b/src/components/TokenSelection/TokenSelection.tsx
@@ -132,6 +132,44 @@ const TokenSelection = ({
     return filterTokens(Object.values(sortedTokens), tokenQuery);
   }, [sortedTokens, tokenQuery]);
 
+  const sortedFilteredTokens: TokenInfo[] = useMemo(() => {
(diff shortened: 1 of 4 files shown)


piersss (approved):
Noice!

piersss (changes requested):
Looks good! I found a small bug:

1. Go to Ethereum main net
2. Remove UNI token from active list
3. Try to re-add it. You can't because all these AAMMUNIAAVEWETH tokens appear on top.

gpxl-dev (approved):
I can't reproduce the bug @piersss but what you are describing is what I was talking about in my previous comment - I think @codyenokida has since fixed it - can you make sure you have pulled and retry?

From my perspective this is working well now :)

Sidenote that uniswap currently has no image url based on our tokenlist - the horrible rendering with the text placeholder will be fixed when we merge to develop, because I replaced it with the `TokenLogo` component that uses `backgroundImage` so won't render the alt text.

Issue #179, opened 2021-08-26
Route tokens instead of content paths


dmosites:
Rather than routing `/#/swap` etc. route from and to tokens e.g. `/#/:tokenFrom/:tokenTo` to support linking and reloads

PR #181, opened 2021-08-26
Added default tokens (USDT, WETH) to SwapWidget

piersss:
I think the default should be USDT -> WETH, but after the user selects a token, those should be saved to local storage and used as the new default. I can add that if you guys agree.

Edit: This ticket will be combined with #179


gpxl-dev (approved):
Excellent :+1:

PR #184, opened 2021-08-27
Fix/no result token query

codyenokida:
Added fix when no token results are found via token query, it falls back to "No results found" text.


piersss (approved):
Works!

gpxl-dev (approved):
Nice! :+1:

PR #193, opened 2021-09-04
Feature/wallet selection redesign

piersss:
Added restyle of wallet selection and removed react-modal from project


gpxl-dev (changes requested):
The contents of the modal are great, but the behaviour should be the same as the `TokenSelection` modal.

My original plan was to use a styled version of react-modal to easily achieve this, but given the `TokenSelection` implementation strayed away from this, we should make our own component that can be re-used to achieve the overlay, transition and back button with custom children.

piersss:
> The contents of the modal are great, but the behaviour should be the same as the `TokenSelection` modal.
> 
> My original plan was to use a styled version of react-modal to easily achieve this, but given the `TokenSelection` implementation strayed away from this, we should make our own component that can be re-used to achieve the overlay, transition and back button with custom children.

Ah yes, I will make an overlay component.

piersss:
I have added the Overlay component.

gpxl-dev (approved):
Looking awesome :ok-hand:

PR #194, opened 2021-09-06
Use on chain data if tokenLists are missing tokens

### LINKED ISSUE #168: Fetch metadata directly from ERC20 contract for tokens that are on the registry but not in token lists
We should be able to add ERC20 tokens by address even if they aren't in our chosen tokenLists by fetching their symbol and decimals directly from the contract.

Are there any gotchas to consider?

gpxl-dev:
Fixes #168

- Waits for tokenLists and supportedTokens to be fetched
- Determines if any supportedTokens don't have TokenInfo from tokenlist
- Fetches missing tokenInfo using `scrapeToken` from `@airswap/metadata`
- Caches metadata in localStorage (`airswap/metadataCache/[chainId]`) and doesn't requests subsequently.

Note that localStorage is automatically freed up on load if either metadata or supportedTokens change such that some previously stored TokenInfo is now extraneous.

At time of PR there are 6 tokens supported in the registry that are not in our Token lists - you can use one of these to test:

```text
0x83f798e925bcd4017eb265844fddabb448f1707d
0x16de59092dae5ccf4a1e6439d611fd0653f0bd01
0xbc5cef436eadacadfa8dafb63088f09f21dea7e9
0xc2cb1040220768554cf699b0d863a3cd4324ce32
0xd6ad7a6750a7593e092a9b218d66c0a814a3436e
0xe6354ed5bc4b393a5aad09f21c46e101e692d447
```

### DIFF
--- a/src/features/metadata/metadataApi.ts
+++ b/src/features/metadata/metadataApi.ts
@@ -1,6 +1,7 @@
-import { fetchTokens } from "@airswap/metadata";
+import { fetchTokens, scrapeToken } from "@airswap/metadata";
 import { TokenInfo } from "@uniswap/token-lists";
 
+import { providers } from "ethers";
 import uniqBy from "lodash.uniqby";
 
 const tokensCache: {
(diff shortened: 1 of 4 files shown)


dmosites:
LGTM. Do we have a default style or icon to use for the scraped tokens?

gpxl-dev:
@dmosites good point, at the minute it's blank - easy to replace, ideally with an image, but can be an icon or whatever.
Probably a question mark on a solid background?

dmosites (approved):
Let's discuss icon handling with designers but think this is good to go for now.

PR #213, opened 2021-10-04
adds norwegian locale

robbestad:
I have taken care to find proper substitions and use correct grammar, so it is not a word-for-word translation. I know it is hard to evalute translations when you are not a native speaker, but maybe there are others in the community that also speaks the language and could provide a review?


dmosites (approved):
Looks good! We can adjust in the future for any needed edits.

PR #222, opened 2021-10-14
Language theme popover

codyenokida:
- added language popover settings 
- localStorage for theme and language
- added temporary hover and focus states
- added system theme button, which looks at system preferences and toggles accordingly

feedback appreciated :D


codyenokida:
Thanks for the comments, I def need to pay attention to details (which makes sense, I am totally not a super detail oriented person). I fixed up to the figma doc, and updated per recommendations. Thanks!

piersss:
Styling still needs a little bit of tweaking. Not blocking so I'll approve.

PR #226, opened 2021-10-20
Reactivate wallet on reload

robbestad:
I think the API may have changed, so a small change to look for the address inside the lastConnectedAddress made it so that the activate process got triggered.

I did make some changes to the packages, as my setup defaulted to typescript 3.4.4 instead of 3.3.5 and that cascaded into a long list of type errors. Also, the wallet connector needed a fixed version as the src code wasn't actually distributed in the latest version it fetched.

### DIFF
--- a/src/features/balances/balancesApi.ts
+++ b/src/features/balances/balancesApi.ts
@@ -3,7 +3,7 @@ import balancesDeploys from "@airswap/balances/deploys.js";
 import { Light, Wrapper } from "@airswap/libraries";
 
 import erc20Abi from "erc-20-abi";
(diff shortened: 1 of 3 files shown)


robbestad:
> what are the ts-ignore comments for? Does it have to do with the version issue you were having? Anyway we can get around the ignores?
> 
> Apart from the above comments, this LGTM.

Thanks for the review! Those ts-ignores are there because I cant start the project otherwise. I made sure I'm on version 4.3.5 of Typescript, but they still persist and prevents the project from compiling. Can you check which version you're running?

Issue #214, opened 2021-10-05
Listen to swap events instead of tracking transactions


gpxl-dev:
Currently (100% RFQ interface), when an order is submitted, we listen to the transaction being successful or being reverted. 

For Last Look, the front end won't have access to the transaction (as the maker will submit the transaction).

Instead, we should listen for swap events with the same nonce, and race it against the expiry.

This will consolidate the logic regardless of the protocol used.

gpxl-dev:
Relevant sections in code to look at (there may be others)!

After taking an order:
https://github.com/airswap/airswap-web/blob/9b9fae360bb97c02d7f6fd73afae70b11fa2adc5/src/features/orders/ordersSlice.ts#L106

After loading the page if there were pending transactions when the user left previously:
https://github.com/airswap/airswap-web/blob/9b9fae360bb97c02d7f6fd73afae70b11fa2adc5/src/features/transactions/Transactions.tsx#L37

robbestad:
I'd be happy to work on it, though I'm sure it would have be synced up with work being done on the Last Look. Am I right in assuming that instead of polling for a transactionreceipt you would want to listen an even that would either return an RFQ swap receipt or a Last Look receipt? Or would you keep the polling and run the event listener on the nonce and resolve whichever returns first?
 
In any case, I would start by setting up a tests for the transaction so that RFQ is tested, and then expand it by adding skeleton tests for Last Look.

gpxl-dev:
@robbestad awesome - feel free to assign yourself.

The way I am thinking about it, I don't think it is dependent on last look. Both RFQ and last look are successful when a swap event with the nonce and correct counterparties is emitted, and both have an expiry that we can use to indicate failure.

This will probably be merged before last look, so the last look branch can pull in the changes before it's completed.

PR #230, opened 2021-10-21
Feature/swap eventlistener

robbestad:
This PR sets up an event listener for swap events when the user connects.
* If the user has an account loaded, the event listener is subscribed to
* If the user disconnects the wallet, the event listener unsubscribes
* If the user connects a wallet, the listener gets subscribe to
* If the user has an active listener but closes or refreshes the page, the listener gets unsubscribed
* And of course, when a swap event occurs, the user gets notified 

This needs to be tested and verified that it meets the goal set up in #214


src/features/wallet/Wallet.tsx
```diff
@@ -83,6 +84,35 @@ export const Wallet: FC<WalletProps> = ({ className = "" }) => {
   const [connector, setConnector] = useState<AbstractConnector>();
   const [provider, setProvider] = useState<WalletProvider>();
   const [activated, setActivated] = useState(false);
+
+  useBeforeunload(() => {
```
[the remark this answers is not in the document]
robbestad: When you're developing with hot reload you're gonna have a lot of subscriptions going if you don't clean up.  It's also very polite :)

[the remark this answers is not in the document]
gpxl-dev: Agree although pretty sure hot reload won't trigger beforeunload, because the page doesn't unload.

gpxl-dev (changes requested):
Nearly there, but a couple of errors. 
I've pushed suggested changes in bc322d74350f530842a1bb430e765b3a4f1e3e4b. Summary below.

I've tested this branch tracking all 3 types of **successful** swap (wrapped rfq, standard rfq, last look) in all conditions (stay on page from start to finish, refresh page between submission and completion, leave page after submission & return after completion). Last Look swaps are stuck at processing if the swap listener isn't connected at the point the swap completes - have made a note on the dev meeting agenda tomorrow to discuss this but propose putting it in a different PR.

(I have **not** yet tested with two instances of the app connected to the same address and I have **not** tested failures)

**`swapEventSubscriber.ts`**

- prefer verbs for function names
- `signerWallet` is the 3rd (index: `2`) argument not the 7th
- added some comments to explain flow in `mapSwapEvent` and tried to make it cleaner. I would argue passing named typed arguments in is easier to read and follow (see [bigswap bot listener code](https://github.com/gpxl-dev/airswap-big-swaps-discord-bot/blob/main/index.ts#L50-L63)) but happy to leave it if you prefer this.
- Don't dispatch anything for transactions we don't know about (i.e. other people's swaps)

**`handleTransaction.ts`**

- re-add the code to periodically check for a transaction we can't find, but only for non "Order" transactions (i.e. approve, deposit, withdraw), because these won't emit a swap and get picked up by the swap listener. (Apologies for the misdirection)


**`transactionActions.ts`**

- make all properties of `mineTransaction` payload optional, because there aren't any that we always have
- never pass the transaction nonce, just the swap nonce if there is one

**Tests**

- Tests were looking for nonce being parsed out from the `_hex` value of a `BigNumber` object. Using values prefixed with `_` when using a library is generally bad practise as the prefix tends to imply that it's internal and should ideally not be used. More likely to break without a breaking semver change. Because the tests relied on the code using this behaviour, they currently fail. Could be fixed by making the `BigNumber` args actually bignumbers in your sample data.
- expectation of `signerWallet` to be walletAccount for RFQs was wrong, but was passing due to `signerWallet` being assigned the value for `senderWallet` (see above)
- Protocol is now pulled from found transaction in the store so doesn't work with the test.

**Misc**

- change `todo` to `TODO:` to be picked up by plugins & parsers.

src/features/transactions/handleTransaction.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @param transactionInState
+ * @param walletHasChanged
+ * @param dispatch
+ * @param library
+ */
+async function handleTransaction(
+  transactionInState: SubmittedTransaction,
+  walletHasChanged: boolean,
+  dispatch: any,
+  library: any
```
piersss, line 17:
```
dispatch: Dispatch<Action>,
library: Web3Provider,
```

src/features/transactions/swapEventSubscriber.ts
```diff
@@ -0,0 +1,144 @@
...
+  chainId,
+  dispatch,
+}: {
+  nonce: string;
+  transactionHash: string;
+  signerWallet: string;
+  transaction: any;
+  protocol: "last-look" | "request-for-quote" | undefined;
+  chainId: number;
+  dispatch: any;
```
piersss, line 31: `Dispatch<Action>`

src/features/transactions/swapEventSubscriber.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    return handleReceipt({
+      nonce,
+      signerWallet,
+      transactionHash,
+      protocol,
+      chainId,
+      transaction,
+      dispatch,
+    });
```
piersss, line 142: This is function is easy to mistype since nonce, signerWallet and transactionHash are all strings. Too bad you can't type on string length.

```
return handleReceipt({
  nonce,
  transactionHash,
  signerWallet,
  transaction,
  protocol,
  chainId,
  dispatch,
})
```

src/features/transactions/swapEventSubscriber.ts
```diff
@@ -0,0 +1,144 @@
...
+  signerWallet,
+  transaction,
+  protocol,
+  chainId,
+  dispatch,
+}: {
+  nonce: string;
+  transactionHash: string;
+  signerWallet: string;
+  transaction: any;
```
piersss, line 28: ```transaction: SubmittedTransaction;```

src/features/transactions/swapEventSubscriber.ts
```diff
@@ -0,0 +1,144 @@
...
+  protocol: "last-look" | "request-for-quote" | undefined;
```
piersss, line 29: `protocol?: "last-look" | "request-for-quote"`

src/features/transactions/swapEventSubscriber.ts
```diff
@@ -0,0 +1,144 @@
...
+  dispatch(
+    mineTransaction({
+      signerWallet: signerWallet,
+      nonce: nonce,
+      hash: transactionHash,
+      protocol,
+    })
+  );
+
+  notifyTransaction("Order", transaction, tokens, false, chainId);
```
piersss, line 44: Modify notifyTransaction to receive transaction SubmittedTransaction type

src/features/transactions/transactionsSlice.ts
```diff
@@ -51,16 +65,28 @@ const initialState: TransactionsState = {
 
 function updateTransaction(
   state: TransactionsState,
+  nonce: string,
   hash: string,
-  status: "processing" | "succeeded" | "reverted"
+  signerWallet: string,
+  status: "processing" | "succeeded" | "reverted",
+  protocol?: "request-for-quote" | "last-look"
```
piersss, line 72: I think we should add types or enums for status and protocol types.

piersss (changes requested):
Tested thoroughly and did not find any bugs. Very nice. Just some minor suggestions in code, that's all!

src/features/transactions/handleTransaction.ts
robbestad: thanks, fixed it in #280

src/features/transactions/swapEventSubscriber.ts
robbestad: Hear ya! Changed in #280

src/features/transactions/swapEventSubscriber.ts
robbestad: Put the fix in #280 :)

src/features/transactions/swapEventSubscriber.ts
robbestad: Put the fix in #280 :)

src/features/transactions/transactionsSlice.ts
robbestad: Agree, got a little tedious typing that in.

PR #258, opened 2021-10-27
Validator integration

### LINKED ISSUE #210: Display any errors from Validator prior to Swap
https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=2037%3A3836

hmalik88:
Closes #210 

Integration of the `Validator` contract into the web app. The `ErrorList` component was created to display the errors returned from our `Validator` contract and is rendered into the `SwapWidget` through the `Overlay` component.

Would appreciate some guidance with the styling as you can see it's not that great ATM, I tried my best to replicate the Figma design.

NOTE: This integration is pre-launch and will be updated with `checkSwap` calls to the `Light` contract, the `Validator` repo is currently deprecated.


src/components/Overlay/Overlay.tsx
```diff
@@ -29,14 +35,20 @@ const Overlay: FC<OverlayProps> = ({
...
 }) => {
   const shouldReduceMotion = useReducedMotion();
 
   return (
     <Container hasTitle={!!title} isHidden={isHidden}>
       <TitleContainer>
-        <StyledTitle type="h2">{title}</StyledTitle>
+        <TitleSubContainer>
+          <StyledTitle type="h2">{title}</StyledTitle>
+          {!!subTitle && !isHidden && (
```
piersss, line 48: Removing it when `isHidden` will warp the close animation. This should be animated with css, just like the title.

src/components/ErrorList/ErrorList.tsx
```diff
@@ -0,0 +1,72 @@
+import { useRef, useEffect, useState } from "react";
+import { useTranslation } from "react-i18next";
+
+import validatorErrors from "../../../public/locales/en/validatorErrors.json";
+import useWindowSize from "../../helpers/useWindowSize";
+import IconError from "../Icon/icons/IconError";
+import { ScrollContainer, Container } from "../TokenList/TokenList.styles";
```
piersss, line 7: You're using styled components from TokenList. We should refactor this and move ScrollContainer and Container to Overlay/Overlay.styles.

piersss (commented):
I think I need to refactor the Overlay component to include scrolling for better reusability. I can make a branch based on your branch, ok?

hmalik88:
> I think I need to refactor the Overlay component to include scrolling for better reusability. I can make a branch based on your branch, ok?

Yeah, that's fine by me.

gpxl-dev (changes requested):
Looks mostly good, a couple of changes I'd suggest:

1. Would prefer less duplication in `LastLook.tsx` - suggest renaming `getCurrentOrder` to `getSignedOrder`, and then changing the arguments for `sendOrderForConsideration` to just a `locator` (`string`) and an `order` (`LightOrder`)
2. The `errors` state variable should probably be explicitly `validatorErrors` or `validationErrors` for clarity. 
3. Is `Validator.ts` going to be in `@airswap/libraries` at some stage? 

I haven't checked through the styling / jsx as I assume piers has covered this already!

hmalik88:
> Looks mostly good, a couple of changes I'd suggest:
> 
> 1. Would prefer less duplication in `LastLook.tsx` - suggest renaming `getCurrentOrder` to `getSignedOrder`, and then changing the arguments for `sendOrderForConsideration` to just a `locator` (`string`) and an `order` (`LightOrder`)
> 2. The `errors` state variable should probably be explicitly `validatorErrors` or `validationErrors` for clarity.
> 3. Is `Validator.ts` going to be in `@airswap/libraries` at some stage?
> 
> I haven't checked through the styling / jsx as I assume piers has covered this already!

Will do to the suggested changes and `Validator.ts` will not be in libraries, next iteration of `Light` will house the `checkSwap` function.

gpxl-dev:
Gotcha thanks!

gpxl-dev (approved):
Looks good - not sure about merging while Validator still shows swaps as unauthorized.

gpxl-dev:
Updated to fix last look swaps, and to skip the validator for wrapper swaps (because of the `senderWallet` balance check - since removed in the version within the new swap contract).

Now works with all 3 types.

codyenokida (approved):

PR #267, opened 2021-11-01
[Feature] New design for transactions tab

codyenokida:
Thanks @Japlito for designing this! 

- Implemented new tab for the transactions tab.
- added ETH counter on the top right of screen
- changed persisting transactions to 3 -> 10.

todos to perhaps tackle on a separate ticket:
1. how to handle transactions with no expiry on the "active transactions" section.
2. Discussion between what's the difference between "Expired" vs. "Failed". just syntactical or is there a fundamental diff between those two which we need to implement?


gpxl-dev (approved):
Looks good! I notice we're retrieving the most recent 10 transactions from our history - wondering if we should have ways to:

- view more than 10
- clear the history

And also if there is a limit to the number that we'd keep? These are outside of the scope of this PR for sure though - just thinking out loud!

robbestad (approved):

PR #270, opened 2021-11-01
Switch chain button

### LINKED ISSUE #153: Improve UX when wallet isn't connected
Currently it's possible to click the token select but token lists aren't populated due to not having a chain ID yet.

### Discussed in https://github.com/airswap/airswap-web/discussions/144

<div type='discussions-op-text'>

<sup>Originally posted by **wjmelements** August 19, 2021</sup>
Can make it work with fallback jsonrpc; otherwise don't show it.</div>

gpxl-dev:
- Disables inputs when wallet not connected and reduces opacity of swap inputs
- Doesn't show the ? in the token select when a wallet isn't connected
- Shows 'not connected' and 'unsupported network' text accordingly
- Checks for the connected wallet being selected on an unsupported chain
- Changes button when appropriate to call the RPC method to prompt MetaMask to switch chain ([EIP-3326](https://ethereum-magicians.org/t/eip-3326-wallet-switchethereumchain)) 
- Removes some extraneous code along the way

Closes #153 

Note: This will only work with injected providers at present (i.e. those setting `window.ethereum`), but this probably covers most users. Have asked in web3-react repo if there is a way to access other providers easily .


piersss (changes requested):
Tested. Switched chains in all states of the app tried to make it crash. Found one:

When I switch chains during a swap (DAI -> WETH), on the quote screen. I get this error:

```
InfoSection.tsx:149 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'symbol')
    at InfoSection (InfoSection.tsx:149)
```

Also another UI thing, is it possible to show the default tokens (USDT -> WETH) in a disconnected state when no tokens are present in window.location?

robbestad (approved):
LGTM

PR #272, opened 2021-11-02
Add reverse ENS lookups for wallet address

gpxl-dev:
Went a little off-piste by doing this but wanted to add it because it was easy!

Updates the wallet component to check for ENS records pointing to the address, and renders the ENS name instead of address if there is one.

To avoid too many queries, we cache in memory.
Only works for the chain on which the ENS name is registered.

Note: one of the rare cases where I found using [useLayoutEffect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) was better, as it avoided a flash when opening and closing the wallet component (which re-renders the `WalletAddress` component)

**UPDATE 3rd Nov** - as the transactions tab adds another place where the wallet address is displayed, I pulled the code out into a hook for easy re-use.


robbestad (approved):
Nice!

codyenokida (commented):
I've refactored the wallet button and transactions tab, can we hold off on this change till after that merge goes in?

gpxl-dev:
@codyenokida yes, was planning to :)

robbestad (approved):

PR #280, opened 2021-11-03
Feature/swap eventlistener rebased

robbestad:
Same as https://github.com/airswap/airswap-web/pull/230 but rebased and merged with recently closed PRs.


src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -487,7 +483,10 @@ const SwapWidget = () => {
       if (e.code && e.code === 4001) {
         // 4001 is metamask user declining transaction sig
       } else {
-        // FIXME: notify user - toast?
+        notifyError({
```
gpxl-dev, line 486: This was not a merge confict, just a bonus!

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -445,13 +445,9 @@ const SwapWidget = () => {
         // Setting quote amount prevents the UI from updating if pricing changes
         dispatch(setTradeTermsQuoteAmount(bestTradeOption!.quoteAmount));
         // Last look order.
-        //@ts-ignore
-        // TODO: give order the correct type
         const { order, senderWallet } = await LastLook.getSignedOrder({
           locator: bestTradeOption!.pricing!.locator,
           terms: { ...tradeTerms, quoteAmount: bestTradeOption!.quoteAmount },
-        }).catch((e) => {
```
gpxl-dev: catching here is wrong - if we can't generate the order we need to hit the catch on line 486. It was also the source of the type error, because `order` could be undefined if `getSignedOrder` throws (the catch returns nothing)

src/features/orders/ordersSlice.ts
```diff
@@ -374,9 +374,7 @@ export const take = createAsyncThunk(
         dispatch(submitTransaction(transaction));
       }
     } catch (e: any) {
-      console.error(e);
-      // TODO: don't throw the actual error message, just show a helpful message
-      dispatch(declineTransaction({ reason: "", hash: "" }));
+      dispatch(declineTransaction(e.message));
```
gpxl-dev, line 377: reverted due to consistency with the rest of the file. Also looked at where this is actually handled and it only logs to the console (hence removing the additional log). AFAIK this is hit if the user cancels the metamask popup, so doesn't need to trigger any UI action.

src/components/SwapWidget/SwapWidget.tsx
robbestad: Yeah... I thought I had removed it previously when you commented on it the last time, but it sort of snuck back.

gpxl-dev (approved):
I think it's looking good!
Would appreciate if someone else also runs through as many scenarios as they can (e.g. all the various swap types) just to double check - have done this myself but have possibly lost track of which things I tested at various merge points!

PR #282, opened 2021-11-03
265: Added mixin file with scrollbar styling

piersss:
- Added a style/mixins.ts file for reusable styles.
- Added ScrollBar mixin and added to multiple components


gpxl-dev (approved):
nice :)

PR #283, opened 2021-11-03
Replace animated background blur with static shadow

gpxl-dev:
- Blur implemented as per designs
- Added `shadows` key to theme, expect to use this in future for other similar effects


piersss (approved):
Why can't we have nice things. :(

PR #286, opened 2021-11-04
Performance improvements

gpxl-dev:
Draft.

So far:
- Only render transaction list when it is open


piersss:
This should be fine. Also fixes the tabbing bug.

PR #310, opened 2021-11-04
Fix order api types

### LINKED ISSUE #227: Missing type of provider will throw error on revert transactions
`orderSlice.ts 70` has `provider` as `any` type. Should be `ethers.providers.Web3Provider`. This causes some refactoring to be needed in the orderSlice. 

Additionally doing this shows that for revert transactions, we are not giving a reason; thus will probably crash the app, since we are passing in a string instead of {hash: string, reason: string} object.

`115 dispatch(revertTransaction(receipt.transactionHash));`

gpxl-dev:
@codyenokida - I looked into how to get the revert reason from a failed transaction. Your google fu may be stronger than mine, but I came to the conclusion there is no reliable way to do it at present for a number of reasons.

Added to the fact we don't actually display the "reason" for a revert anywhere currently, I hard coded "transaction reverted" as the reason. Let me know if you think there's a better option.

The other changes were straightforward.

Fixes #227

### DIFF
--- a/src/components/SwapWidget/SwapWidget.tsx
+++ b/src/components/SwapWidget/SwapWidget.tsx
@@ -504,7 +504,7 @@ const SwapWidget = () => {
           chainId: chainId!,
           senderAmount: baseAmount,
           senderTokenDecimals: baseTokenInfo!.decimals,
-          provider: library,
+          provider: library!,
         })
       );
       await unwrapResult(result);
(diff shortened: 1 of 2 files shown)


robbestad (approved):

codyenokida (approved):
Yep, sounds about right. LGTM!

PR #312, opened 2021-11-04
Highlight wallet pill when transactions pending

### LINKED ISSUE #308: Pending tx: connected wallet pill should pulse / glow while there is 1 or more pending transaction

### LINKED ISSUE #325: Widget glow is currently brighter than the designs
See:
https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=4278%3A7334

gpxl-dev:
Fixes #308, Fixes #325

Biggest part of this PR is subjective about whether the animation is ok/correct. Note that this uses a pseudo element with a box shadow that is being translated around underneath the button element so performance is not an issue as it was with the blur filter on the SwapWidget.

Here's a video. The glow also fades in when the transaction starts, but the toast and metamask popup block those most of the time...

An alternative could be a simple pulse (i.e. scale up and down on loop), but personally I liked the movement.

https://user-images.githubusercontent.com/82473383/140434055-8262defc-c7c6-4ff5-84e3-a5091302b706.mov


piersss (approved):
This works great. Metamask disappears just in time for the glow animation to start. Just the toast covers it fading out, but that's fine.

I think we should also highlight the pending transaction in your wallet when you open it.

robbestad (approved):
LGTM

robbestad:
It does look a bit exaggarated on light mode

codyenokida (approved):
wow this is so neat. looks good to me :)

PR #314, opened 2021-11-05
Fix/285 buttons hover focus outline

### LINKED ISSUE #285: Make sure all buttons have correct border / outline.
According to jop all buttons with borders need to be darkGrey on hover. Right now outline is the same but maybe we need to differentiate from hover to comply with UI standards.

### LINKED ISSUE #305: Update current bidirectional arrow on input form screen to be single arrow pointing from top to bottom

piersss:
Added a InputOrButtonBorderStyle and BorderlessButtonStyle mixin that is used on several buttons and links.

- Closes #285 
- Closes #305 because the icon was really bothering me.
- Also removed the old airswap icon from project


robbestad (approved):
LGTM

codyenokida (approved):

PR #320, opened 2021-11-05
modifies transaction key

robbestad:
Not all transactions have a hash, so those end up key-less (with an error in the console). This fix concats the hash with a nonce so that they'll have a unique key. Vice versa, not all transactions have a nonce, so you can't rely on that either. But they all have either one.


gpxl-dev (approved):
Pretty sure this is fine as per our discord sanity check!

PR #323, opened 2021-11-05
Fix clearing app state on resize

### LINKED ISSUE #322: App state resets on window resize

gpxl-dev:
Closes #322 

This was a memoization problem - the resize triggers the `useWindowSize` hook to update the state in `App.tsx` which in turn rerenders all of the child components, including the `LastLookProvider` which was recomputing the value passed to the context, and therefore triggering the hook to clear the state on chain change that required the `LastLook` context as a dependency because it needs it to disconnect from all servers.


piersss (approved):
Nice, was wondering why that happened.

PR #328, opened 2021-11-05
Remove bookmark warning

### LINKED ISSUE #327: Remove bookmark warning from app

gpxl-dev:
Closes #327


piersss (changes requested):
Also remove from userSettingsSlice.ts

robbestad:
Why not just redesign the component so that it fits in with the new design instead of removing all handling?

gpxl-dev:
@robbestad - think it's just because of availability of design resources, plus based on discord chat we're not actually sure if it's required.

Because of the nature of git, this stuff isn't lost, we can just revert the commit if/when we're ready to re-implement?

My original approach was to temporarily remove it but keep logic, but I guess that's a good way to collect legacy code that is no longer used!

robbestad (approved):

PR #330, opened 2021-11-07
Fix/264 redesign quote screen

### LINKED ISSUE #264: Redesign of quote screen.
- The coin icons will be switched around.
- Extra buttons and modals
- When user is eligible fee is reduced.
- "New quote" timer has moved to the top

https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=4278%3A5853

Screenshot 2021-10-30 at 16 08 14

Screenshot 2021-10-30 at 16 08 02

Screenshot 2021-10-30 at 16 07 37

Screenshot 2021-10-30 at 16 07 29

piersss:
Closes #264 

- Redesigned quote screen
- Added new transition animation
- Added new modals

TODO (in separate ticket):
- Determine if user has AST staked for fee text.

https://user-images.githubusercontent.com/86911296/140638998-efb3f338-f0ff-4921-b551-766705a4a673.mov

### DIFF
--- a/public/locales/en/information.json
+++ b/public/locales/en/information.json
@@ -5,5 +5,18 @@
     "getStarted": "Get started",
     "paragraph": "Protocol fees are automatically distributed to members for contributing to product and community development.",
     "paragraph2": "Check out the following guides to get started as a..."
+  },
+  "gasFreeSwaps": {
+    "title": "Gas free swaps",
+    "paragraph": "AirSwap introduces gas-free swapping!",
+    "paragraph2": "With this protocol you will not need to pay any gas to swap!"
+  },
+  "protocolFeeDiscount": {
+    "title": "Protocol fee discount",
+    "paragraph": "AirSwap community members get a discount when trading on AirSwap by staking AST.",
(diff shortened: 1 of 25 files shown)


hmalik88 (changes requested):
Quote timer is showing after an RFQ swap is performed. Other then that, this looks really good!

Screen Shot 2021-11-08 at 2 26 32 AM

robbestad (approved):
Agree with @hmalik88 observation. Otherwise, looks good!

piersss:
> Quote timer is showing after an RFQ swap is performed. Other then that, this looks really good!
> 
> Screen Shot 2021-11-08 at 2 26 32 AM

Cool, nice find!

codyenokida (approved):
piers, you are a beast. This looks awesome. One comment about it is animating the signer token after you click continue. But that's super minor and pretty subjective.

hmalik88 (approved):

piersss:
> piers, you are a beast. This looks awesome. One comment about it is animating the signer token after you click continue. But that's super minor and pretty subjective.

Thanks! Yeah my intention was to animate the signer token as well. But current state also works. The ETH -> WETH goes as intended btw. Not sure! We can change it later if we want. :)

Issue #303, opened 2021-11-04
Set mouse cursor to "denied" on input form when wallet not connected


dmosites:

gpxl-dev:
it's also possible to tab to the inputs even when they're "disabled". Bad implementation, sorry!

PR #331, opened 2021-11-08
Set mouse cursor to denied on input form when wallet not connected

robbestad:
Fix for #303 (also tabindex for tabbing to input field)


hmalik88 (approved):
LGTM.

Issue #287, opened 2021-11-04
Transactions drawer should slide "in" rather than "over" and be able to be kept open


dmosites:
See: https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=5621%3A6416

robbestad:
should also make "untabbable" so that you can't use the tab button to go to it without opening it.

PR #339, opened 2021-11-09
[Fix] Transactions tab slide in

codyenokida:
- changed transactions tab styling
- One bug: blur background is rerenders when swap widget moves position

https://user-images.githubusercontent.com/30182368/140841948-b14ec274-6e6b-46f7-bcb5-6de8e7291386.mov

Are there any recommendations to fix this?


piersss (changes requested):
Hey cody. Transition looks great. Not sure how we can fix the glow, maybe we should use a png as background. It's barely noticeable on my screen though.

I checked half of the code since I ran out of time, will finish later.

robbestad:
On firefox the gear icon is partially hidden
bilde

robbestad:
I think you can remove the event listeners at https://github.com/airswap/airswap-web/blob/3166085a598d0c0b0876b6ebe7c3ae3aae815caa/src/components/TransactionsTab/TransactionsTab.tsx#L101 if it's supposed to be kept open

src/components/TransactionsTab/TransactionsTab.tsx
```diff
@@ -109,86 +129,121 @@ const TransactionsTab = ({
...
+            </WalletAnchorTag>
+          </WalletHeader>
+
+          <TransactionsContainer
+            ref={transactionsScrollRef}
+            $overflow={overflow}
+          >
             <Legend>
-              <LegendLine>{t("wallet:completedTransactions")}</LegendLine>
+              <LegendLine>{t("wallet:activeTransactions")}</LegendLine>
```
piersss, line 188: Fix the translation for activeTransactions and completedTransactions to "Active" and "Completed", and use style `text-transform: uppercase` in css.

codyenokida:
closes #287

src/components/Icon/icons/IconChevronRight.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { SvgIconProps } from "../Icon";
+
+const IconChevronRight: FC<SvgIconProps> = ({
+  className = "",
+}): ReactElement => (
+  <svg viewBox="0 0 24 24">
+    <path
+      d="M9.99997 6L8.58997 7.41L13.17 12L8.58997 16.59L9.99997 18L16 12L9.99997 6Z"
+      fill="white"
```
gpxl-dev, line 11: I think this should be `currentcolor` which means that the fill is the same as the CSS text colour. This makes it easier to change the colour of the icons (e.g. in light mode). P.S. I learned this from @piersss' code, thanks!

src/components/Icon/icons/IconChevronRight.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { FC, ReactElement } from "react";
+
+import { SvgIconProps } from "../Icon";
+
+const IconChevronRight: FC<SvgIconProps> = ({
+  className = "",
```
gpxl-dev, line 6: `className` isn't used - for other icons, `className={className}` is added to the `svg` elements, so I did this to mine for consistency!

src/components/TransactionsTab/TransactionsTab.tsx
```diff
@@ -109,86 +116,123 @@ const TransactionsTab = ({
...
-                .slice(0, 10)
-                .map((transaction) => (
+            <TransactionContainer>
+              {pendingTransactions.length ? (
+                pendingTransactions.map((transaction) => (
                   <WalletTransaction
                     transaction={transaction}
                     tokens={tokens}
                     chainId={chainId!}
                     key={transaction.hash}
```
gpxl-dev, line 184: not all transacitons in this list will have a hash. Suggest:

```
{`${transaction.hash}-${transaction.nonce}-${transaction.expiry}`}
```

Same for below.

gpxl-dev (changes requested):
Looks great - three minor comments plus a change needed for short ENS names manifesting in the issue reported by Sven on discord.

gpxl-dev (approved):
Looks awesome!

PR #365, opened 2021-11-10
Add progress bars to transactions with expiries

### LINKED ISSUE #321: Add progress bars for each swap (based on expiry) in the transactions widget
See design:
https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=4559%3A6070

* For approvals and WETH transactions second line is a "pending" label.
* Add 20 second buffer to expiry timeout to catch Swap events that might be emitted after the expiry.

gpxl-dev:
Closes #321 

- adds progress bar
- adds the icon from the designs
- removes transaction link ([ref](https://discord.com/channels/590643190281928738/844220258386575371/908071460298457139))

image

### DIFF
--- a/src/components/ProgressBar/ProgressBar.tsx
+++ b/src/components/ProgressBar/ProgressBar.tsx
@@ -1,0 +1,24 @@
+import { FC } from "react";
+
+import { Track, Progress } from "./ProgressBar.styles";
+
+const ProgressBar: FC<{
+  startTime: number;
+  endTime: number;
+}> = ({ startTime, endTime }) => {
+  const now = Date.now();
+  const done = now - startTime;
+  const left = endTime - startTime;
+  const fractionDone = done / left;
+  return (
+    <Track>
+      <Progress
+        animate={{ scaleX: 1 }}
+        initial={{ scaleX: fractionDone }}
+        transition={{ ease: "linear", duration: left / 1000 }}
+      ></Progress>
+    </Track>
+  );
+};
+
+export default ProgressBar;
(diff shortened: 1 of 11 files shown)


src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const done = now - startTime;
+  const left = endTime - startTime;
+  const fractionDone = done / left;
+  return (
+    <Track>
+      <Progress
+        animate={{ scaleX: 1 }}
+        initial={{ scaleX: fractionDone }}
+        transition={{ ease: "linear", duration: left / 1000 }}
+      ></Progress>
```
robbestad, line 19: Both this and the <span></span> below in src/components/TransactionsTab/subcomponents/WalletTransaction/WalletTransaction.tsx can be collapsed as their bodies are empty.

robbestad (dismissed):
Smashing!

gpxl-dev:
Hijacked this branch to add an additional fix for our approvals check (f458981)

- We now wait for supportedTokens to resolve before checking active token balances. This fixes a bug on the first page load on a new browser (because localStorage persists them otherwise) whereby all tokens will request allowance
- To be extra sure, we also check to see if we have actually received an allowance figure in the allowance check. Previously we were checking something equivalent to: `if (!allowances[token]) requiresApproval = true` - instead we now set requiresApproval false in this case to prevent the user approving something they've already approved. We additionally set the error flag to tell the user to refresh the page. This should no longer happen though thanks to the change above and the changes last night in combination.

PR #366, opened 2021-11-10
Prevent showing approve button unneccessarily

### LINKED ISSUE #349: investigate possible allowance failure
Seen in console, but not sure how to reproduce.

```
Error fetching allowances.light: call revert exception (method="walletAllowances(address,address,address[])", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.5.0)
```

Need to check error handling on approval checks as the implication would be asking users to pay for approval unnecessarily (costing gas).

--

* P1: check how many tokens are being requested at once. might be too many args in the balances call.
* P3: initially: if on the quote screen user is about to take the order and allowances call has failed or not yet returned, then display prompt to reload?

gpxl-dev:
Closes #349 

In rare cases, the call to fetch ERC20 token allowances for our contracts may fail. This is a failure that could cause a user to waste funds on gas re-approving tokens that are already approved.

This PR provides an interim fix by not allowing the user to proceed following a failed allowances request. The SwapWidget becomes disabled, and the Info text informs the user that fetching token allowances has failed, and that they'll need to reload the page. The action button becomes a "Reload Page" button.

The best way to test this is to short circuit the allowance check by throwing an error on line 77 of `balancesSlice.tsx`.

As a bonus, this PR also fixes the fact that balances and allowances are requested three times on every page load because new `connector` and `provider` objects are created on every render of `Wallet.tsx`. This may improve efficiency elsewhere in the app too.

### DIFF
--- a/public/locales/en/balances.json
+++ b/public/locales/en/balances.json
@@ -2,5 +2,7 @@
   "tokenBalances": "Token balances",
   "addToTokenSet": "Import",
   "unsupportedToken": "Not yet supported",
-  "balance": "Balance"
+  "balance": "Balance",
+  "failedToFetchAllowances": "Failed to fetch token allowances",
+  "failedToFetchAllowancesCta": "Please reload the page"
 }
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -24,5 +24,6 @@
   "vote": "vote",
   "build": "build",
   "about": "about",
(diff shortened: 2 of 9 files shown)


dmosites (approved):
Tested locally looks good!

codyenokida (approved):

PR #367, opened 2021-11-10
Social links changes

### LINKED ISSUE #350: Social links quick fixes
- Update DOM structure not to have `a` tag within `a` tag.
- Open links in new tab
- go behind the widget
- the horizontal divider is still being displayed on buttons that don't expand
- focus / active of a multi-lingual menu should be on the the menu item instead of the menu

gpxl-dev:
- Fixes link nested in link (caused console error in best case)
- Removes divider that was appearing on hovering even if there were no locales
- Adds a black background on hover to cover the case that the widget is being overlapped by the locales list

Doesn't address the active state when clicking a link (which I think is pretty minor) but otherwise closes #350 - @piersss if you have a good way to deal with the focus problem then please feel free to add it to this PR.

Finally mobile experience needs work here as it's hard to get to the French twitter - not sure of an easy way around this - ideally want to tap once to open the locales list on mobile, but maintain behaviour on desktop.

### DIFF
--- a/src/components/SocialButtons/SocialButtons.styles.tsx
+++ b/src/components/SocialButtons/SocialButtons.styles.tsx
@@ -43,6 +43,9 @@ export const SocialButton = styled.div<{ showLocales: number }>`
   justify-content: flex-end;
   width: 2.5rem;
   height: ${({ showLocales }) => (1 + showLocales) * 2.5 + "rem"};
+  &:hover {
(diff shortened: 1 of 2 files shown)


piersss (approved):
I tried to fix the hover and focus states but I need more time to come up with a good solution. I stashed my changes and will continue later. Let's approve for now since it's not a big issue.

robbestad (approved):

PR #372, opened 2021-11-11
Fix fee calculations

### LINKED ISSUE #362: Last look orders currently don't include swap fee
Reported by aaaaaaaaa on discord, using rinkeby.

- swapping 0.99 WETH -> DAI
- no approval button
- gets following:
image

This is logged just prior to sending to validator (`SwapWidget.tsx:486`)

```
{
  "order": {
    "expiry": "1636558383",
    "nonce": "1636558263281",
    "senderToken": "0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea",
    "senderAmount": "983888000000000000",
    "signerWallet": "0xdb3Ec7B16Fd60fB4fDB58A438Bd8AF57d8d3a91c",
    "signerToken": "0xc778417e063141139fce010982780140aa0cd5ab",
    "signerAmount": "990823630000000000",
    "r": "0xf449e44830a26e462d7536ffd09a3d27f60a7775170e3cec8d199f5974750614",
    "s": "0x4a5bea25edce054e71ad9902fa1b8bf786f6bbd818a43b21adec2b839c73b5c2",
    "v": "28"
  },
  "senderWallet": "0x8e811410CE01e0244808Af95bbA906b8aB77a40b",
  "allowances": {
    "status": "idle",
    "lastFetch": 1636558194830,
    "inFlightFetchTokens": null,
    "values": {
      "0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea": "0",
      "0xc778417e063141139fce010982780140aa0cd5ab": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
      "0xcc1cbd4f67cceb7c001bd4adf98451237a193ff8": "0",
      "0xd9ba894e0097f8cc2bbc9d24d308b98e36dc6d02": "0",
      "0x0000000000000000000000000000000000000000": "0"
    }
  }
}
```

gpxl-dev:
Fixes #362 

The issues:

1. When last look orders were being generated, the entered amount was being used for the `signerAmount`, instead of `signerAmount + fee`. This would result in failed swaps if a user's max balance was used.
2. Fee calculations were using fees that were one order of magnitude too big (0.7% instead of 0.07%). This resulted in quote amounts that were too small.

To reproduce the issues (before checking out this branch):

1. Try to swap your max WETH for DAI on rinkeby. You'll hit the validator error saying you have insufficient balance (note the text is wrong!)
2. Swap less than your max balance and look at the actual amounts sent and received.

To verify this works, checkout the branch and swap your max WETH balance for DAI on rinkeby. You should be left with exactly zero WETH, and receive exactly the amount quoted.

Also updates the WalletTransaction to include `signerFee` in the from amount for last look swaps.

### DIFF
--- a/public/locales/en/validatorErrors.json
+++ b/public/locales/en/validatorErrors.json
@@ -8,7 +8,7 @@
   "invalid_sig": "Invalid signature",
   "expiry_passed": "Order has expired",
   "unauthorized": "Signer has not authorized the signatory",
-  "signer_allowance_low": "Signer has not approved the signer amount",
+  "signer_allowance_low": "Signer does not have enough balance",
   "signer_balance_low": "Signer has not approved the signer amount",
   "nonce_aleady_used": "Nonce has been already used or cancelled",
   "unableSwap": "Unable to swap",
--- a/src/components/TransactionsTab/subcomponents/WalletTransaction/WalletTransaction.tsx
+++ b/src/components/TransactionsTab/subcomponents/WalletTransaction/WalletTransaction.tsx
@@ -4,6 +4,8 @@ import { findTokenByAddress } from "@airswap/metadata";
 import { TokenInfo } from "@airswap/types";
 import { formatUnits } from "@ethersproject/units";
 
+import BigNumber from "bignumber.js";
+
 import {
   SubmittedApproval,
   SubmittedOrder,
@@ -76,6 +78,16 @@ const WalletTransaction = ({
       tokens,
       chainId
     );
+
+    // For last look transactions, the user has sent the signer amount plus
+    // the fee:
+    let signerAmountWithFee: string | null = null;
+    if (tx.protocol === "last-look") {
+      signerAmountWithFee = new BigNumber(tx.order.signerAmount)
+        .multipliedBy(1.0007)
+        .integerValue(BigNumber.ROUND_FLOOR)
+        .toString();
+    }
     return (
       <Container>
         <TextContainer>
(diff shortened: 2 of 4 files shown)


hmalik88 (approved):
LGTM

robbestad (approved):

dmosites:
> note the text is wrong!
Do you know whether it's wrong text, the wrong error key, or the wrong error being returned by the validator? cc @hmalik88

dmosites:
Great catch @gpxl-dev. Rather than handling specific values suggest the three instances of `1.0007` be calculated from `signerFee`. Just need to be divided by `10000` (in the contracts we have a constant called `FEE_DIVISOR`)

hmalik88:
> > note the text is wrong!
> 
> Do you know whether it's wrong text, the wrong error key, or the wrong error being returned by the validator? cc @hmalik88

@dmosites It was the text, pushed a fix. By the way, I know we did away with the sender allowance/balance checks because it's redundant. In this case, for last look, this "signer" balance error refers to the user themselves. Do we want to somehow remove this error for last look orders?

gpxl-dev:
@hmalik88:

> In this case, for last look, this "signer" balance error refers to the user themselves. Do we want to conditionally render this order for only RFQ?

I think it's still useful to render for last look. We shouldn't hit this problem if everything else is working properly, but this will help greatly with support if we do.

gpxl-dev:
@dmosites extracted #375 in interest of time & priorities as I think hard coding is safe for the short term. LMK if you disagree.

PR #376, opened 2021-11-11
[Feature] Temporary mobile screen

codyenokida:
Super makeshift starter for the mobile landing page. 

- please feel free to hijack this pr!


gpxl-dev (changes requested):
Generally good but seems to be visible the whole time if you just scroll down the page?

Also wonder if we should also consider height in the media query - at the moment if I turn my old iPhone 7 to landscape I can see the page but it's not really usable (also I'm on a non web3 browser)

image

gpxl-dev (approved):

PR #383, opened 2021-11-11
Job small fixes and disable light mode

piersss:


robbestad:
as mentioned on Discord, if you want to disable language until the settings is ready, just comment out  `.use(LanguageDetector)` in `src/i18n/i18n.ts`

robbestad (approved):

PR #401, opened 2021-11-12
Include gas prices in quote selection

gpxl-dev:
- API methods:
  - [x] Get gas price in ETH
  - [x] Get token price in WETH from uniswap
  - [x] Get gas price in tokens
- [x] Integrate into order flow


robbestad:
The code looks sane 👍 Just wondering about the results I get (for eth->eth on mainnet) 
bilde

src/features/gasCost/gasCostSlice.ts
```diff
@@ -0,0 +1,75 @@
...
+      }>
+    ) => {
+      state.tokenPrices[
+        action.payload.tokenAddress.toLowerCase()
+      ] = action.payload.tokenPriceInWeth.toString();
+    },
+  },
+  extraReducers: (builder) => {
+    builder
+      .addCase(setWalletDisconnected, (state) => initialState)
```
robbestad, line 49: remove unused `state` from the parantheses

robbestad (approved):

robbestad:
Aha, got it! So, the NaN is the results you get when it has an non-existing or invalid API key. After adding a valid key, I got the proper results.

PR #408, opened 2021-11-18
Adding listener for deposit and withdrawals

### LINKED ISSUE #379: Add listeners for `Deposit` and `Withdrawal` events from WETH contract to catch ETH/WETH "swaps"
https://github.com/makerdao/sai/blob/master/src/weth9.sol#L25

robbestad:
Add listeners for Deposit and Withdrawal events from WETH contract to catch ETH/WETH "swaps"

Events are emitted on` ETH->WETH` and `WETH<-ETH` (and only these are the ones I've tested).

Closes #379

### DIFF
--- a/src/features/transactions/wrapEventSubscriber.ts
+++ b/src/features/transactions/wrapEventSubscriber.ts
@@ -8,14 +8,15 @@ import {
   decrementBalanceBy,
   incrementBalanceBy,
 } from "../balances/balancesSlice";
-import { SubmittedTransaction } from "./transactionsSlice";
+import { SubmittedTransactionWithOrder } from "./transactionsSlice";
 
 const handleWrapEvent = (data: any, dispatch: any) => {
   const transactions = store.getState().transactions;
 
-  const transaction: SubmittedTransaction | null =
-    transactions.all.find((t: any) => t.hash === data[2].transactionHash) ||
-    null;
+  const transaction: SubmittedTransactionWithOrder | null =
+    (transactions.all.find(
+      (t: any) => t.hash === data[2].transactionHash
+    ) as SubmittedTransactionWithOrder) || null;
 
   // If we don't have a 'transaction', we don't already know about this swap
   // and therefore don't need to update the UI.
@@ -23,17 +24,13 @@ const handleWrapEvent = (data: any, dispatch: any) => {
 
   dispatch(
     decrementBalanceBy({
-      //@ts-ignore
       tokenAddress: transaction.order.senderToken,
-      //@ts-ignore
       amount: transaction.order.senderAmount,
     })
   );
   dispatch(
     incrementBalanceBy({
-      //@ts-ignore
       tokenAddress: transaction.order.signerToken,
-      //@ts-ignore
       amount: transaction.order.senderAmount,
     })
   );
--- a/src/features/transactions/wrapEventSubscriber.ts
+++ b/src/features/transactions/wrapEventSubscriber.ts
@@ -10,10 +10,10 @@ import {
 } from "../balances/balancesSlice";
 import { SubmittedTransaction } from "./transactionsSlice";
 
-const handleWrapEvent = (eventName: string, data: any, dispatch: any) => {
+const handleWrapEvent = (data: any, dispatch: any) => {
   const transactions = store.getState().transactions;
 
-  let transaction: SubmittedTransaction | null =
+  const transaction: SubmittedTransaction | null =
     transactions.all.find((t: any) => t.hash === data[2].transactionHash) ||
     null;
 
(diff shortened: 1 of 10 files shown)


src/features/transactions/wrapEventSubscriber.ts
```diff
@@ -0,0 +1,53 @@
...
+      tokenAddress: transaction.order.signerToken,
+      //@ts-ignore
+      amount: transaction.order.senderAmount,
+    })
+  );
+};
+
+export default function subscribeToWrapEvents(params: {
+  wrapContract: Contract;
+  library: any;
```
piersss, line 43: `library: Web3Provider`

src/features/transactions/wrapEventSubscriber.ts
```diff
@@ -0,0 +1,53 @@
...
+  dispatch: Dispatch;
+}) {
+  const { wrapContract, dispatch } = params;
+  wrapContract.on("Deposit", async (...data) =>
```
piersss, line 47: `wrapContract.on("Deposit", async (data) =>`

src/features/transactions/wrapEventSubscriber.ts
```diff
@@ -0,0 +1,53 @@
...
+    handleWrapEvent("Deposit", data, dispatch)
+  );
+  wrapContract.on("Withdrawal", async (...data) =>
```
piersss, line 50: `wrapContract.on("Deposit", async (data) =>`

src/features/wallet/Wallet.tsx
```diff
@@ -114,13 +126,23 @@ export const Wallet: FC<WalletPropsType> = ({
         chainId,
         dispatch,
       });
+      subscribeToWrapEvents({
+        wrapContract,
+        //@ts-ignore
```
piersss, line 131: Removing @ts-ignore works fine for me

piersss (changes requested):
Did some small tests, works well. Noticed small thing:

- Transaction appears in transaction list but no progress bar (or is this ok?)

Will do some more testing later, need to run some errands.

src/features/transactions/wrapEventSubscriber.ts
robbestad: this is a problem, because if I don't spread it I don't get the data I want. All I get is an address
```Deposit 0xa3863513D0f2ECeDcd25ef3b5C73D67cb0aaD6B2```
vs 
```
Deposit 
Array(3) [ "0x5411c75107B2546b404A84ADBD555CBe3eAf4fE0", {…}, {…} ]
```

**Comment by robbestad** [comment]
Regarding this:
* Transaction appears in transaction list but no progress bar (or is this ok?)

These don't have an expiry, so they won't show a progress bar, and I think that's correct.

File: src/features/transactions/wrapEventSubscriber.ts [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+import { Contract } from "ethers";
+
+import { store } from "../../app/store";
+import {
+  decrementBalanceBy,
+  incrementBalanceBy,
+} from "../balances/balancesSlice";
+import { SubmittedTransaction } from "./transactionsSlice";
+
+const handleWrapEvent = (eventName: string, data: any, dispatch: any) => {
```
piersss, line 13: `const handleWrapEvent = (data: any, dispatch: Dispatch<Action>) => {`

eventName is not used, so it can be omitted.

src/features/transactions/wrapEventSubscriber.ts
```diff
@@ -0,0 +1,54 @@
...
+  const transactions = store.getState().transactions;
+
+  let transaction: SubmittedTransaction | null =
```
piersss, line 16: `const transaction: SubmittedTransaction | null =`

src/features/transactions/wrapEventSubscriber.ts
```diff
@@ -0,0 +1,54 @@
...
+    null;
+
+  // If we don't have a 'transaction', we don't already know about this swap
+  // and therefore don't need to update the UI.
+  if (!transaction) return;
+
+  dispatch(
+    decrementBalanceBy({
+      //@ts-ignore
+      tokenAddress: transaction.order.senderToken,
```
piersss, line 27: A SubmittedTransaction is not supposed to have an order, I don't understand. Or is this something beyond the scope of this ticket?

src/features/transactions/wrapEventSubscriber.ts
piersss: Gotcha!

piersss (changes requested):
Works as intended, just some minor issues.

src/features/transactions/wrapEventSubscriber.ts
robbestad: Okay! I used that for logging, but no point in keeping it around

src/features/transactions/wrapEventSubscriber.ts
robbestad: Yeah... this is a weird one that should have its own ticket. If you look at the reducer, most transactions will have orders, and this is where I find the data I need to update the balances. So either the SubmittedTransaction needs to be extended or we must do some other refactoring to avoid using ts-ignore all the time when dealing with this.

robbestad: I did it, let me know what you think

piersss: Yeah, I don't think `SubmittedTransactionWithOrder` is even necessary. I assume all "succeeded" `SubmittedTransaction` should have an order. But I understand it's beyond the scope of this ticket.

piersss (approved):

PR #430, opened 2021-11-20
293: Added reset button

### LINKED ISSUE #293: Clicking on logo should reset app state to initial screen

piersss:
Closes #293 

- Turned airswap logo into reset button
- Added resetOrders action in orders store

### DIFF
--- a/src/components/SwapWidget/SwapWidget.tsx
+++ b/src/components/SwapWidget/SwapWidget.tsx
@@ -118,7 +118,7 @@ const SwapWidget: FC<SwapWidgetPropsType> = ({
   const balances = useAppSelector(selectBalances);
   const allowances = useAppSelector(selectAllowances);
   const bestRfqOrder = useAppSelector(selectBestOrder);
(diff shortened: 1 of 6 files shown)


src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -242,6 +256,10 @@ const SwapWidget: FC<SwapWidgetPropsType> = ({
     );
   }, [allowances.light.status, allowances.wrapper.status]);
 
+  useEffect(() => {
```
robbestad, line 259: I don't think we need this out there

robbestad (changes requested):
Works pretty well and it's very welcome. I noticed it didn't take you back when you're at the "Transaction Submitted" stage. Not a very big deal, since you have a button in the middle of the screen to take you to a "new swap".

src/components/SwapWidget/SwapWidget.tsx
piersss: Removed!

robbestad (approved):

PR #434, opened 2021-11-23
Feature/poeditor integration

robbestad:
This PR changes the i81n setup and usages to match the data structure from Poeditor. There are some places I haven't updated the types for (or couldn't easily find) so I would appreciate some eyes on those. Aside from that, the code is now ready to sync with Poeditor.


piersss (approved):
Great stuff!

Issue #373, opened 2021-11-11
Wrong text for signer balance low subtext on validator


gpxl-dev:
Says something about allowance.

francisldn:
@gpxl-dev, a bit more guidance on this issue would be appreciated, thanks.

gpxl-dev:
Hi @francisldn this should be a very easy one to resolve - looks like it's just swapping the values for the two keys here:

https://github.com/airswap/airswap-web/blob/develop/public/locales/en/translation.json#L138-L139

You can see that the value for "signer_balance_low" mentions allowance (approval), and the allowance text talks about balance.

PR #456, opened 2022-01-08
Corrected text for signer balance low

francisldn:
#373 Corrected the wrong text for signer balance low under validator errors.

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -135,8 +135,8 @@
     "invalid_sig": "Invalid signature",
     "expiry_passed": "Order has expired",
     "unauthorized": "Signer has not authorized the signatory",
-    "signer_allowance_low": "Signer does not have enough balance",
-    "signer_balance_low": "Signer has not approved the signer amount",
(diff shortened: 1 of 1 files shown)


francisldn:
Hi @gpxl-dev, it's an easy fix, but would appreciate if you could please review and approve this PR.

piersss:
Hey Francis, I'm not sure what the correct translation it, so lets's wait for @gpxl-dev. Do you have access to the poeditor? All translations changes should be added in there as well. Contact @robbestad or @dmosites for an invite.

piersss:
> #373 Corrected the wrong text for signer balance low under validator errors.

_Fixes #373_ will link the issue.

gpxl-dev (approved):

gpxl-dev:
@piersss its possible/likely that this is an English locale only issue as the values for these two keys were the wrong way around

robbestad:
> @piersss its possible/likely that this is an English locale only issue as the values for these two keys were the wrong way around

agree

PR #459, opened 2022-01-25
Migrate to ethgas.watch API for gas prices

gpxl-dev:
This PR changes the source of gas prices (used to compare Last Look and RFQ swaps) from defipulse to ethgas.watch

Experience has shown ethgas.watch has proven to be much more reliable, and it's also simpler to use because there's no API key required.


dmosites (approved):
LGTM

PR #462, opened 2022-01-28
Issue 380 - /whitepaper route added as static HTML

Phasor:
Had a few CSS conflict issues with the previous PR for this issue. Fixed on this branch.


src/components/Toolbar/Toolbar.tsx
```diff
@@ -70,6 +70,11 @@ const Toolbar: FC<ToolbarProps> = ({ onButtonClick }) => {
           text={t("common.join")}
           onClick={() => onToolbarButtonClick("join")}
         />
+        <ToolbarButton
```
piersss, line 73: Use react-router for internal links in the app. So modify ToolbarButton so it accepts link as property and use react-router accordingly.

src/components/Whitepaper/Whitepaper.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { FC } from "react";
+
+import htmlContent from "./WhitepaperContent";
+
+const Whitepaper: FC = () => {
+  return (
+    <>
+      <div dangerouslySetInnerHTML={{ __html: htmlContent }} />
+    </>
```
piersss, line 9:
```
return (
  <div dangerouslySetInnerHTML={{ __html: htmlContent }} />
);
```

src/components/Toolbar/Toolbar.tsx
```diff
@@ -70,6 +70,11 @@ const Toolbar: FC<ToolbarProps> = ({ onButtonClick }) => {
...
+          iconName="learn"
+          text="Whitepaper"
```
piersss, line 75: Add a translation for whitepaper with `useTranslation`. Ask Don or Brigand to help you add it to poeditor.

Also the word whitepaper doesn't fit. Please contact design how to solve this.
Screenshot 2022-01-31 at 16 53 54

piersss (changes requested):
Hey again phasor!

I guess we could use pure css instead of styled components for this if the other developers are ok with it.

One nitpicky thing: could you use camelCase for the files in /public?

Also the toolbar was not designed for 7 buttons, so it doesn't fit anymore. But that is something beyond the scope of this ticket so I could fix that later.

src/components/Toolbar/Toolbar.tsx
piersss: No need for toolbar button anymore. So this is resolved.

src/components/Toolbar/Toolbar.tsx
piersss: No need for toolbar button anymore. So this is resolved.

src/components/Whitepaper/Whitepaper.tsx
Phasor: Updated thanks

piersss (approved):
All good!

Issue #461, opened 2022-01-28
Change text of Metamask button when Metamask is not installed.


piersss:
Right now it just fails silently. Solution:

The button should say: _Get Metamask_ and links to https://metamask.io

GiuntaLucas:
@piersss the development is already made by myself, i'm just waiting to have the access to the repo to push my branch, move de ticket and create my pr.

piersss:
> @piersss the development is already made by myself, i'm just waiting to have the access to the repo to push my branch, move de ticket and create my pr.

Ah! No problem. Will update ticket.

PR #466, opened 2022-02-06
Feature/461 metamask button

GiuntaLucas:
- Change the text from the button when Metamask is not installed
- Redirect to the metamask web site on click on the button in this case

Ticket: https://github.com/airswap/airswap-web/issues/461

get metamask


piersss (changes requested):
Hello LeHardy.

Welcome to the team!

I see a lot of changes. But the only file that should be changed is `WalletProviderList`, I think your editor did some formatting. Please revert these changes.

piersss (approved):

Issue #433, opened 2021-11-22
Add support for XDEFI wallet: https://www.xdefi.io/


dmosites:

GiuntaLucas:
I check the documentation from xDefi it's seems not possible to use Metamask and xDefi at the same time.
If the user decide to use xDefi by default, they will override the ethereum object to act like metamask.
They are working on that to be able to use both on the same dapp.

So for the moment it's Metamask OR xDefi. To be able to do the switch, the user has to enable/disable the option "Prioritise XDEFI⚠️" from the wallet.

PR #467, opened 2022-02-06
Feature/433 xdefi wallet

GiuntaLucas:
- Add support of xDefi
- Follow the same logic of metamask when xDefi is not installed (change text + redirect to xDefi web page)

Ticket: https://github.com/airswap/airswap-web/issues/433

Warning: xDefi act like Metamask, when metamask is already installed, metamask will be open instead of xDefi until the user will activate an option from his xDefi wallet to be priority.

get metamask
connect


piersss (changes requested):
Hey LeHardy

I never used xdefi wallet so I will install it later this week and test things out.

This MR has a lot of overlap with the metamask button MR. So you can use the same feedback here.

piersss:
Steps I took:
1. Open incognito browser with xdefi as only extension
2. Enabled "Prioritise XDEFI" (or else it won't connect anyway)
3. Open web app and connect wallet
4. It injects the wallet but the registry tokens are not filled

After connecting the wallet the registry tokens are not filled. It fails silently in registryApi.ts, line 20:

```
  const [addEvents, removeEvents] = await Promise.all([
    registryContract.queryFilter(addTokensEventFilter),
    registryContract.queryFilter(removeTokensEventFilter),
  ]);
```

Wrapping it around a try catch I get this error:

```
base-provider.ts:1607 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')
    at Web3Provider.<anonymous> (base-provider.ts:1607:1)
    at Generator.next (<anonymous>)
    at fulfilled (base-provider.ts:2:1)
```

My guess is that xdefi is not fully compatible with some ethers.js functions. This is a tough one.

robbestad (changes requested):
It's necessary to rebase this to the current `develop` and go from there. I also concur with the feedback left from Pierss

dmosites:
Closing until there's a clear path froward.

Issue #423, opened 2021-11-19
Make WalletConnect fit in Overlay component.


piersss:

GiuntaLucas:
i took a look into the WalletConnect from our dependency Web3-react, seems we can't change the modal via this lib. They are working into the v8 wich will give a big update of the lib but still in beta now. What we can do for the moment is to point the css class into our globalCss and apply what we want. I took the opportunity to fix the issue from the icon and the WalletConnect wich is a bit wrong and the color of the "mobile" and "desktop".

piersss:
> i took a look into the WalletConnect from our dependency Web3-react, seems we can't change the modal via this lib. They are working into the v8 wich will give a big update of the lib but still in beta now. What we can do for the moment is to point the css class into our globalCss and apply what we want. I took the opportunity to fix the issue from the icon and the WalletConnect wich is a bit wrong and the color of the "mobile" and "desktop".

I'm afraid it's gonna take a while before v8 is released. Let's use your solution for now then.

PR #458, opened 2022-01-20
Feature/457 mobile design

piersss:
Closes #457 

Main changes:
- Removed Temporary landing page
- Fully responsive app for desktop, tablet and mobile
- New top bar for mobile and tablet portrait (<768px)
- New mobile toolbar menu on mobile and tablet portrait (<768px)
- Transactions list is now a full screen menu on mobile
- New mobile menu in transactions list with "View on etherscan", "Copy address", and "Disconnect" buttons
- Added mobile and desktop sizes in typography

https://user-images.githubusercontent.com/86911296/150287047-8915f6eb-30cf-4eb2-a51a-105ca90433fb.mov

Original design under "Latest version":

https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/New-AirSwap-Website?node-id=77%3A1014

Tested it on android metamask app. Anyone with iphone who could test?

PR #468, opened 2022-02-06
feature/423 walletconnect fit overlay

GiuntaLucas:
- Fix the issue from the dialog header about the icon and the label
- Fix the color from the buttons 'QR Codes' and 'Desktop' from the dialog

Ticket: https://github.com/airswap/airswap-web/issues/423

Capture

### DIFF
--- a/src/style/GlobalStyle.ts
+++ b/src/style/GlobalStyle.ts
@@ -43,6 +43,13 @@ const GlobalStyle = createGlobalStyle`
     font-weight: ${(props) => props.theme.typography.paragraph.fontWeight};
     line-height: ${(props) => props.theme.typography.paragraph.lineHeight};
   }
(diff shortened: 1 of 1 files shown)


piersss (approved):
Hey LeHardy!

Easy and simple fix. Hopefully they will update their library, it also doesn't work well on mobile.

GiuntaLucas:
Thank you @piersss . Before to initiate the merge, i see you are speaking about the fact the walletconnect doesnt work from mobile, just to be sure, the airswap web app is not yet available for mobile phone right ? I only have access to the redirection screen (vote, discord, blog,  twitter ...)

piersss:
We're working on it! :)

https://github.com/airswap/airswap-web/pull/458

PR #469, opened 2022-02-06
Feature/185 search engine optimizations

### LINKED ISSUE #185: Search engine optimization
- [ ] Add noindex for static, locales, etc.
- [ ] Work on sitemap / figure out URLs
- [ ] Review attached report

piersss:
Fixes #185 

Let me know what you think!

Main changes:

- Added whitepaper and join to sitemap.xml
- Added pages folder (the 2 pages are Home and Whitepaper now)
- Join modal is now directly accessibly via /join
- Added react-helmet so we can have different meta tags for pages
- Replaced HashRouter with BrowserRouter because why do we need the /#/ in url's?
- Swap title is now h1

A question I still have is this: right now the title is just "AirSwap". I came up with "AirSwap - Open source web app for instant swaps" for the time being.


robbestad:
I like the suggestion for the title. It adds a lot of value. I think you should even add the chains supported for swapping.  "AirSwap - Open source app for gas free swaps on Ethereum" for instance

src/App.tsx
```diff
@@ -1,65 +1,41 @@
...
-            </Web3ReactProvider>
-            <GlobalStyle />
-          </ThemeProvider>
-        </Route>
-      </Switch>
-    </Router>
+    <>
+      <HelmetContainer
+        title={"AirSwap - Open source web app for instant swaps"}
+        description={"AirSwap powers peer-to-peer trading."}
```
robbestad, line 20: Adding Ethereum here would help with SEO. Also gas free instead of instant? I don't know if instant brings any value, seeing as every other swap service are just as fast.
Furthermore, what do you think of localization? Probably a bit of a hassle if it doesn't work out of the box with the locale function

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -863,7 +877,7 @@ const SwapWidget: FC<SwapWidgetPropsType> = ({
       <Overlay
         title={t("validatorErrors.unableSwap")}
         subTitle={t("validatorErrors.swapFail")}
-        onClose={async () => {
+        onCloseButtonClick={async () => {
```
robbestad, line 880: I would consider using "useCallback" for async buttons.

robbestad (changes requested):
Wrt localization I'm not sure it can be resolved easily, but let me know what you think

piersss:
> I like the suggestion for the title. It adds a lot of value. I think you should even add the chains supported for swapping. "AirSwap - Open source app for gas free swaps on Ethereum" for instance

So if we release on Avalanche it would say: "AirSwap - Open source app for gas free swaps on Ethereum and Avalanche"?

piersss:
> Wrt localization I'm not sure it can be resolved easily, but let me know what you think

Answered here: https://github.com/airswap/airswap-web/pull/469#discussion_r803748088 . What do you think?

src/components/SwapWidget/SwapWidget.tsx
piersss: Not written by me, but I see what I can do. :)

PR #470, opened 2022-02-09
added new static sitemap

Phasor:
In relation to issue 465 - sitemaps.

We only have 2 static urls that we would want a search engine to index. With this in mind, a simple approach might be best?

I added a new XML sitemap to /sitemap.xml. The only thing I don't like it that the lastmod field is not dynamic. Do we want to think about how we do that or are we happy with the static values as is?


piersss (commented):
You can close this. There's already a PR for this:

https://github.com/airswap/airswap-web/pull/469

PR #474, opened 2022-02-10
Fixes /whitepaper route and adds helmet to whitepaper component

Phasor:


src/pages/Whitepaper/whitepaperContent.tsx
```diff
@@ -5,8 +5,8 @@ const htmlContent = `<!DOCTYPE html>
     <!-- Required meta tags -->
```
piersss, line 5: I just noticed whitepaperContent has html body and head tags, they shouldn't be in there.

piersss (changes requested):
Nice!

src/pages/Whitepaper/whitepaperContent.tsx
Phasor: Good spot. Updated below.

src/pages/Whitepaper/whitepaperContent.tsx
```diff
@@ -539,7 +502,6 @@ cancel(&lt;makerAddress&gt;, 5, 'GNO', &lt;takerAddress&gt;, 10, 'BAT', &lt;expi
       ga('create', 'UA-98809379-1', 'auto')
```
piersss, line 502: What about all these scripts? Are they necessary?

piersss (changes requested):

piersss:
I have added the commits to my branch, so I will close this PR.

PR #476, opened 2022-02-21
Wraps call in try/catch to avoid crash in dev

### LINKED ISSUE #464: Fix New Mobile Site Bug  - WETH to ETH
On Rinkeby, when I swap WETH to ETH I get the following bug:

image
image

robbestad:
Closes #464

### DIFF
--- a/src/hooks/useReferencePriceSubscriber.ts
+++ b/src/hooks/useReferencePriceSubscriber.ts
@@ -27,9 +27,15 @@ const useGasPriceSubscriber = () => {
     () => () => {
       if (intervals.current.gas) return;
       const updateGasPrice = async () => {
-        const price = await getFastGasPrice();
-        if (price) {
-          dispatch(setFastGasPrice(price.toString()));
+        let price = "0";
+        try {
+          let _price = await getFastGasPrice();
+          if (_price) price = _price.toString();
+        } catch (e) {
+          console.error(e);
+        }
+        if (parseFloat(price)!==0) {
+          dispatch(setFastGasPrice(price));
         }
       };
       intervals.current.gas = window.setInterval(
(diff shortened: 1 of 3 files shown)


src/hooks/useReferencePriceSubscriber.ts
```diff
@@ -27,9 +27,15 @@ const useGasPriceSubscriber = () => {
...
-          dispatch(setFastGasPrice(price.toString()));
+        let price = "0";
+        try {
+          let _price = await getFastGasPrice();
+          if (_price) price = _price.toString();
+        } catch (e) {
+          console.error(e);
+        }
+        if (parseFloat(price)!==0) {
+          dispatch(setFastGasPrice(price));
```
dmosites, line 38: Figure this can be moved up to line 33 (replacing `price = _price.toString()`)

robbestad: Certainly. It looks much better that way

src/hooks/useReferencePriceSubscriber.ts
```diff
@@ -27,9 +27,13 @@ const useGasPriceSubscriber = () => {
...
-        const price = await getFastGasPrice();
-        if (price) {
-          dispatch(setFastGasPrice(price.toString()));
+        try {
+          let price = await getFastGasPrice();
+          if (price) {
+            dispatch(setFastGasPrice(price.toString()));
+          }
+        } catch (e) {
+          console.error(e);
```
piersss, line 36: How about adding an extra function with comment to make it a little bit more readable?

```
// Fix for call not working in dev
const getFastGasPriceFix = async (): Promise<string> => {
  try {
    const price = await getFastGasPrice();
    return price ? price.toString() : "0";
  } catch (e) {
    console.error(e);
    return "0";
  }
}

-----

const updateGasPrice = async () => {
 const price = await getFastGasPriceFix();
 if (parseFloat(price)!==0) {
    dispatch(setFastGasPrice(price));
 }
};
```

piersss: Ok never mind!

piersss (approved):

PR #494, opened 2022-02-24
adds title and meta description tags to /whitepaper.html

Phasor:
For seo.

### DIFF
--- a/public/whitepaper.htm
+++ b/public/whitepaper.htm
@@ -1,26 +1,28 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-    <title>Airswap Whitepaper - Peer to peer decentralized exchange</title>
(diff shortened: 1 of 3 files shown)


public/whitepaper.htm
```diff
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+    <title>Airswap Whitepaper - Peer to peer decentralized exchange</title>
+    <meta name="description" content="Learn about the peer to peer decentralized swap protocol from Airswap.">
```
robbestad, line 5: I think it should be identical to the others: "AirSwap: Peer-to-peer Token Trading DEX and Open Source Developer DAO"

robbestad (changes requested):
Good call to add a title and description. :)

PR #495, opened 2022-02-25
Updates deploys.js contract addresses to v3

Phasor:
Only V3 Rinkeby address is in here. Needs updating when we add more chains.


dmosites:
Closing in favor of #496

PR #502, opened 2022-03-18
Polygon support

dmosites:


piersss:
Works on matic and matic testnet mumbai, but no trading pairs!

robbestad (approved):

PR #505, opened 2022-03-27
Feature/503 metamask errors

### LINKED ISSUE #503: Show MetaMask errors to user
Right now errors from MetaMask are only shown in console and a "Error submitting swap" toast. We should better inform the user what error is exactly occurring. Here's a list of possible errors:

https://github.com/MetaMask/eth-rpc-errors/blob/main/src/error-constants.ts

Questions for Don:

1. Should a MetaMask error appear in a toaster or in the Error overlay component?
2. Should translations be added for each error?
3. Only fix the most common "User rejected transaction" error. (quick fix)

piersss:
Fixes #503 

- Shows errors from MetaMask to user in the ErrorList component
- User rejection of trade will just show a toaster
- Added an error file where all errors are sourced from
- Fixed translation typo


src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -604,14 +620,7 @@ const SwapWidget: FC<SwapWidgetPropsType> = ({
         dispatch(clearTradeTermsQuoteAmount());
       }
 
-      if (e.code && e.code === 4001) {
```
piersss, line 607: This didn't work anymore. Because the error gets "rethrown" the `code` gets moved into the `stack`.

dmosites (approved):

PR #519, opened 2022-04-22
506: Replace 0x000..000 with native token in url

### LINKED ISSUE #506: Replace `0x000...000` with `native` (or comparable) for better looking URLs
https://www.airswap.io/#/swap/native/0x27054b13b1b798b345b591a4d22e6562d47ea75a

piersss:
Fixes #506 

I used constants/nativeCurrency.ts to get the name of native token. For example, an url could be:
- `swap/eth/0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea`
- `swap/0xc778417e063141139fce010982780140aa0cd5ab/matic`


dmosites:
Would this require the network to be changed on page load (based on URL) or does the alias in the URL just update based on currently selected network?

piersss:
Fixed!

dmosites (approved):

PR #525, opened 2022-04-29
513: Added getProgressBarAnimationProps helper and tests

### LINKED ISSUE #513: Swap progress bar not working correctly
In some cases have seen the progress bar in a "complete" state even though the swap is being mined. It should progress from 0 to 100% based on the expiry time specified on the order.

Screen Shot 2022-04-13 at 5 29 06 PM

piersss:
Fixes #513 

There were 2 problems:
- Logic for calculating time left and progress was not working.
- Framer-motion not reacting well to restarting animations

So I rewrote logic and added helper with some tests and did animations with simple css.

### DIFF
--- a/src/components/ProgressBar/ProgressBar.styles.tsx
+++ b/src/components/ProgressBar/ProgressBar.styles.tsx
@@ -1,4 +1,3 @@
-import { motion } from "framer-motion";
 import styled from "styled-components/macro";
 
 export const Track = styled.div`
@@ -9,9 +8,15 @@ export const Track = styled.div`
   overflow: hidden;
 `;
 
-export const Progress = styled(motion.div)`
+interface ProgressBarProps {
+  initialWidth: number;
+  duration: number;
+}
+
+export const Progress = styled.div<ProgressBarProps>`
   height: 100%;
-  width: 100%;
+  width: ${(props) => props.initialWidth}%;
   background-color: ${(props) => props.theme.colors.primary};
   transform-origin: left;
+  transition: width ${(props) => props.duration}s linear;
 `;
--- a/src/components/ProgressBar/ProgressBar.styles.tsx
+++ b/src/components/ProgressBar/ProgressBar.styles.tsx
@@ -15,8 +15,9 @@ interface ProgressBarProps {
 
 export const Progress = styled.div<ProgressBarProps>`
   height: 100%;
-  width: ${(props) => props.initialWidth}%;
+  transform: scaleX(${({ initialWidth }) => initialWidth});
   background-color: ${(props) => props.theme.colors.primary};
   transform-origin: left;
-  transition: width ${(props) => props.duration}s linear;
+  transition: transform ${({ duration }) => duration}s linear;
+  will-change: transform;
 `;
--- a/src/components/ProgressBar/ProgressBar.tsx
+++ b/src/components/ProgressBar/ProgressBar.tsx
@@ -1,21 +1,30 @@
-import { FC } from "react";
+import { FC, useEffect, useState } from "react";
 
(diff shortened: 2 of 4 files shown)


src/components/ProgressBar/ProgressBar.styles.tsx
```diff
@@ -9,9 +8,15 @@ export const Track = styled.div`
...
-export const Progress = styled(motion.div)`
+interface ProgressBarProps {
+  initialWidth: number;
+  duration: number;
+}
+
+export const Progress = styled.div<ProgressBarProps>`
   height: 100%;
-  width: 100%;
+  width: ${(props) => props.initialWidth}%;
```
Mitsichury, line 18: could use desctructuration:
`
${({initialWidth}) =>initialWidth }%;
`

src/components/ProgressBar/ProgressBar.styles.tsx
```diff
@@ -9,9 +8,15 @@ export const Track = styled.div`
...
   background-color: ${(props) => props.theme.colors.primary};
   transform-origin: left;
+  transition: width ${(props) => props.duration}s linear;
```
Mitsichury, line 21: could use destructuration:
`
${({duration}) =>duration}%;
`

gpxl-dev: I haven't checked this recently, but in the past, at least, transitioning width was always less performant than transitioning `transform` and changing the `scaleX`. I think it's because transitioning width can trigger a reflow/layout wheras transforms only affect compositing.

I'm not 100% sure if this is still the case in 2022, but it's fairly simple to use `transform: scale` instead of width here, right?

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -1,21 +1,30 @@
...
-  const done = now - startTime;
-  const left = endTime - startTime;
-  const fractionDone = done / left;
+  const [isMounted, setIsMounted] = useState(false);
+
+  useEffect(() => {
+    setIsMounted(true);
+  }, []);
+
+  const { progress, duration } = getProgressBarAnimationProps(
```
gpxl-dev, line 16: naming here confused me initially - I thought `progress` would update. Wonder if this might be easier to understand:

```typescript
const { progress: initialProgress , duration } = getProgressBarAnimationProps(
```

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -1,21 +1,30 @@
...
   endTime: number;
 }> = ({ startTime, endTime }) => {
-  const now = Date.now();
-  const done = now - startTime;
-  const left = endTime - startTime;
-  const fractionDone = done / left;
+  const [isMounted, setIsMounted] = useState(false);
+
+  useEffect(() => {
+    setIsMounted(true);
```
gpxl-dev, line 13: Took me a minute to understand why this was needed. Might be worth adding a comment e.g.:

```
// `isMounted` will be false until the component has rendered once. We use this
// so that we can set the initial width of the progress bar to correspond with
// the initial progress, before setting the "target" to 100% for the CSS transition
```

gpxl-dev (approved):
Looks good! Some optional comments but happy for merge - I'm probably just being slow.

src/components/ProgressBar/ProgressBar.tsx
piersss: Yeah, initialProgress would be more fitting.

src/components/ProgressBar/ProgressBar.tsx
piersss: Sharp!

PR #530, opened 2022-05-25
Fix potential race condition causing bad UI state

### LINKED ISSUE #521: Bad state on main widget
(mainnet, running locally)
1. Completed a USDC/ETH swap
2. Clicked "new swap" after that'd completed
3. Selected the amount field and found myself in this state:

Screen Shot 2022-04-26 at 1 37 37 PM

Active orders might not have been cleared after clicking "new swap"

gpxl-dev:
Fixes #521 

Copied from my response in the issue itself:

> I've checked out the code quite extensively and the only way I can see this could happen is a kind of race condition:
>
> - Orders are re-requested when we get to within 1M of the exiry
> - User clicks "new swap"
> - Response from makers comes in and populates the orders state before the user has a chance to do anything. Once there's an order in the orders state the SwapWidget changes its behaviour and could cause the state displayed in the screenshot.
>
> To mitigate this I'll do a couple of things:
>
> - Cancel the re-request when take is successfully fulfilled (it's currently only being cancelled when we click "new swap")
> - Don't update the orders slice with orders if its state wasn't "requesting".

Note that I couldn't reproduce this issue, but theoretically you could do so if you are running a maker by artificially introducing a long delay before responding, and ensuring you click "New swap" just after the request has gone out

### DIFF
--- a/src/features/orders/orderApi.ts
+++ b/src/features/orders/orderApi.ts
@@ -139,7 +139,7 @@ export function orderSortingFunction(a: Order, b: Order) {
   }
   // If tokens transferred are the same
   if (a.signerAmount === b.signerAmount && a.senderAmount === b.senderAmount) {
-    return aTimeToExpiry - bTimeToExpiry;
+    return parseInt(b.expiry) - parseInt(a.expiry);
   }
   if (a.signerAmount === b.signerAmount) {
     // Likely senderSide
--- a/src/features/orders/orderApi.ts
+++ b/src/features/orders/orderApi.ts
@@ -15,6 +15,8 @@ import {
   utils,
 } from "ethers";
 
+import { RFQ_EXPIRY_BUFFER_MS } from "../../constants/configParams";
+
 const REQUEST_ORDER_TIMEOUT_MS = 5000;
 
 const erc20Interface = new ethers.utils.Interface(erc20Abi);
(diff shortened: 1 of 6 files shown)


piersss:
> 



> Fixes #521
> 
> Copied from my response in the issue itself:
> 
> > I've checked out the code quite extensively and the only way I can see this could happen is a kind of race condition:
> > 
> > * Orders are re-requested when we get to within 1M of the exiry
> > * User clicks "new swap"
> > * Response from makers comes in and populates the orders state before the user has a chance to do anything. Once there's an order in the orders state the SwapWidget changes its behaviour and could cause the state displayed in the screenshot.
> > 
> > To mitigate this I'll do a couple of things:
> > 
> > * Cancel the re-request when take is successfully fulfilled (it's currently only being cancelled when we click "new swap")
> > * Don't update the orders slice with orders if its state wasn't "requesting".
> 
> Note that I couldn't reproduce this issue, but theoretically you could do so if you are running a maker by artificially introducing a long delay before responding, and ensuring you click "New swap" just after the request has gone out

I reproduced the error by setting `timeTilReRequest` in `ordersSlice.ts` line 294 to something really low like 10. This would never happen unless the maker or client date is out of sync. But it would be saver if this value was checked if it's above a certain value or at least positive.

src/features/orders/ordersSlice.ts
```diff
@@ -291,8 +294,17 @@ export const request = createAsyncThunk(
...
       const bestOrder = [...orders].sort(orderSortingFunction)[0];
+      const now = Date.now();
       const expiry = parseInt(bestOrder.expiry) * 1000;
-      const timeTilReRequest = expiry - Date.now() - RFQ_EXPIRY_BUFFER_MS;
+      // Due to the sorting in orderSorting function, these orders will be at
+      // the bottom of the list, so if the best one has a very short expiry
+      // so do all the others. Return an empty order array as none are viable.
+      if (expiry - now < RFQ_EXPIRY_BUFFER_MS) return [];
+
+      const timeTilReRequest = Math.min(
```
piersss, line 304:
```suggestion
      const timeTilReRequest = Math.max(
```

With `Math.min` `RFQ_MINIMUM_REREQUEST_DELAY_MS` acts like a maximum here.

piersss (changes requested):
Small issue. I will see if I can fix the tests.

src/features/orders/orderApi.ts
```diff
@@ -118,9 +120,26 @@ export async function takeOrder(
...
+  } else if (
+    bTimeToExpiry < RFQ_EXPIRY_BUFFER_MS &&
+    aTimeToExpiry >= RFQ_EXPIRY_BUFFER_MS
+  ) {
+    return -1; // prefer A
+  }
   // If tokens transferred are the same
   if (a.signerAmount === b.signerAmount && a.senderAmount === b.senderAmount) {
-    return parseInt(b.expiry) - parseInt(a.expiry);
+    return aTimeToExpiry - bTimeToExpiry;
```
piersss, line 142: This fixed the following ordersSlice test:

```
should select orders based on longest expiry if token amounts are equal
```

src/features/orders/ordersSlice.ts
gpxl-dev: good catch!

src/features/orders/orderApi.ts
gpxl-dev: Ah, it's making the tests pass but not for the right reasons. We pass strings for expiries in the tests so this is definitely needed.

gpxl-dev:
@piersss I've fixed this now. 
Jest types were wrong and thought it would be nicer to use the built in to mock & unmock the date.

piersss (approved):
All good!

PR #534, opened 2022-05-29
463: Added social buttons to mobile menu

### LINKED ISSUE #463: Social buttons for mobile
Social buttons are now hidden on mobile. Mobile design should be updated accordingly.

piersss:
Fixes #463

### DIFF
--- a/src/components/Page/Page.styles.tsx
+++ b/src/components/Page/Page.styles.tsx
@@ -48,23 +48,23 @@ export const StyledPage = styled.div`
 `;
 
 export const StyledSocialButtons = styled(SocialButtons)`
-  display: flex;
-  position: fixed;
-  bottom: 1.5rem;
-  right: 1.5rem;
+  display: none;
 
-  @media ${breakPoints.phoneOnly}, ${breakPoints.shallowScreenOnly} {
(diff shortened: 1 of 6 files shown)


GiuntaLucas (approved):
It's working perfectly. But i have a small question, What do you think to place the button at the bottom of the screen ? As a user experience, maybe it will be better to not see the share button at the same lvl of the menu, and to avoid an overlay when you touch the social button with some locale. 

image

piersss:
> It's working perfectly. But i have a small question, What do you think to place the button at the bottom of the screen ? As a user experience, maybe it will be better to not see the share button at the same lvl of the menu, and to avoid an overlay when you touch the social button with some locale.
> 
> image

Yup, you're right. Will add this.

GiuntaLucas (approved):
Impecable ! Perferct !

2xAST:
In safari on iPhone the buttons are below the url bar, so they are not visible and you can't scroll down to them

piersss:
> In safari on iPhone the buttons are below the url bar, so they are not visible and you can't scroll down to them

Whoops, same on android. Maybe aligning it to the bottom wasn't so smart. 🧐

PR #547, opened 2022-06-09
feature/issue-534

GiuntaLucas:
Display the import default list when no tokens are already imported.

### DIFF
--- a/src/components/TokenList/TokenList.tsx
+++ b/src/components/TokenList/TokenList.tsx
@@ -183,17 +183,16 @@ const TokenList = ({
                 )}
               </TokenContainer>
             )}
-            {inactiveTokens.length !== 0 &&
-              sortedFilteredTokens.length < 5 && (
-                <InactiveTokensList
-                  inactiveTokens={inactiveTokens}
-                  supportedTokenAddresses={supportedTokenAddresses}
-                  onTokenClick={(tokenAddress) => {
-                    addActiveToken(tokenAddress);
-                    setTokenQuery("");
-                  }}
-                />
-              )}
+            {inactiveTokens.length !== 0  && (
+              <InactiveTokensList
+                inactiveTokens={inactiveTokens}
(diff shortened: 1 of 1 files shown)


src/components/TokenList/TokenList.tsx
```diff
@@ -184,7 +184,6 @@ const TokenList = ({
               </TokenContainer>
             )}
             {inactiveTokens.length !== 0 &&
-              tokenQuery &&
               sortedFilteredTokens.length < 5 && (
```
piersss, line 187:
```suggestion

```

This line should also be removed or else the tokens still won't appear if the `sortedFilteredTokens` are more than 5. All else seems fine!

piersss (changes requested):

src/components/TokenList/TokenList.tsx
GiuntaLucas: I removed this condition but i have another question: Do we still show to the user only 10 token from the inactive one by default or do we have to increase the list ? Maybe we can increase to 50 - 100, what do you think ? 

image

piersss: Not sure, did you ask around in the dev chat yet?

piersss (approved):

src/components/TokenList/TokenList.tsx
```diff
@@ -107,10 +107,8 @@ const TokenList = ({
 
   // only take the top 10 tokens
   const inactiveTokens: TokenInfo[] = useMemo(() => {
-    return filterTokens(Object.values(sortedInactiveTokens), tokenQuery!).slice(
-      0,
-      10
-    );
+    const tokens = filterTokens(Object.values(sortedInactiveTokens), tokenQuery!);
+    return tokens.length >= 100 ? tokens.slice(0, 100) : tokens;
```
piersss, line 111: `return filterTokens(Object.values(sortedInactiveTokens), tokenQuery!).slice(
      0,
      10
    );
`

PR #551, opened 2022-06-19
Feature/504 native token fee

### LINKED ISSUE #504: When clicking max ETH, subtract 0.01 gas fees.

piersss:
Fixes #504 

Subtract a transaction fee when native token is used as "from' amount. This amount can be set for any chain in the `nativeCurrencySafeTransactionFee` variable. For example main net is now set to 0.01 and test nets are set to 0.001.

Tooltip is shown when info label is active (desktop and mobile). Jop helped me out with design:
Screenshot 2022-06-19 at 16 45 09
Screenshot 2022-06-19 at 16 45 26


piersss:
> LGTM. Language should make clear that it's to cover the EVM transaction fee and not an additional protocol fee.

Yeah, that's a tough one! Maybe use "gas fee" instead of "transaction fee"?

GiuntaLucas:
Just run the branch in local everything is working fine. Quick question, what's the behavior for the others network ? I don't have any token to try to others network, but the tooltip has to be display with the information regarding to each network ? (MATIC for Polygon, ...)

GiuntaLucas (commented):
Just run the branch in local everything is working fine. Quick question, what's the behavior for the others network ? I don't have any token to try to others network, but the tooltip has to be display with the information regarding to each network ? (MATIC for Polygon, ...)

piersss:
If it's not set in `nativeCurrencySafeTransactionFee` then there will be no deduction and no tooltip will be shown.

gpxl-dev (approved):
Looks good!

We need to populate the amounts for the other networks we're deployed on, right? Will mention this in the Discord.

PR #577, opened 2022-07-09
feature/565 dropdown component

coltraneyan:
The JSX is formatted as divs rather than an option/select module to reduce the amount of boilerplate from the W3 example.

the label, name props were adjusted accordingly:

interface {
category: SelectOption,
options: SelectOption[],
onChange: (option: SelectOption) => void
}

Important note on the styling:

Will need to add styled-component props which allow for adjusting the dimensions of the component, as from my testing, they must have hardcoded width and height values in order to operate correctly. It's siblings in the parent element will share these traits so that it can interact as if it is in-flow with the parent, as the position: relative/absolute currently remove it from that state. Themes are also hardcoded at the moment, I will change these once I dig into the styled-component Theming the app has set up already. 

This took a lot of experimenting--let me know if I overlooked anything, as I changed a lot from the initial example in order for it to mostly follow the figma set up.

### DIFF
--- a/src/components/Dropdown/Dropdown.styles.tsx
+++ b/src/components/Dropdown/Dropdown.styles.tsx
@@ -1,0 +1,97 @@
+import styled from "styled-components";
+
+// theme, fonts, and dimensions are currently hardcoded
+// needs to take in a prop with the relevant dimensions from the parent
+
+export const Item = styled.div`
+  display: none;
+  position: relative;
+  text-align: center;
+  height: 25px;
+  width: 100px;
+  padding-left: 5px;
+
+  &.selected {
+    display: flex;
+    background-color: black;
+    border: 1px solid grey;
+    border-right: none;
+    align-items: center;
+  }
+`;
+
+export const Arrow = styled.div`
+  position: relative;
+  border: 1px solid grey;
+  border-radius: 0 10px 10px 0;
+  border-left: none;
+  height: 25px;
+  padding-left: 5px;
+  padding-right: 5px;
+  display: flex;
+  align-items: center;
+`;
+
+export const Selector = styled.div`
+  display: block;
+  cursor: pointer;
+  position: relative;
+  width: fit-content;
+  height: fit-content;
+  border-radius: 5px;
+  z-index: 1;
+
+  &.show > ${Item} {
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <Wrapper>
+      <AbsoluteWrapper
+        onClick={(event) => {
+          event.currentTarget.classList.toggle("show");
+        }}
+      >
+        <Selector>
+          <Item
+            className={`${
+              selectedOption === category ? "selected category" : "category"
```
piersss, line 45: No reason to show `'selected category'`, `selectedOption` is required so just show its `label`.

piersss (changes requested):
Hey coltrane, welcome to the team! When you make a PR push your branch to the airswap repo next time.

I placed the Dropdown in the SwapWidget header and it seems your styling is not same as figma design. Please double check everything.

Screenshot 2022-07-10 at 13 06 21

Screenshot 2022-07-10 at 13 05 13

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,69 @@
...
+export type SelectOption = {
+  label: string;
+  value: string;
+};
+
+export type DropdownProps = {
+  category: SelectOption;
+  options: SelectOption[];
+  onChange: (option: SelectOption) => void;
+};
```
piersss, line 21:
```suggestion
  className?: string;
};
```

If you want the parent to style this component a `className` should be passed to your `Wrapper` component.

src/components/Dropdown/Dropdown.styles.tsx
```diff
@@ -0,0 +1,97 @@
...
+  border-radius: 0 10px 10px 0;
+  border-left: none;
+  height: 25px;
+  padding-left: 5px;
+  padding-right: 5px;
+  display: flex;
+  align-items: center;
+`;
+
+export const Selector = styled.div`
```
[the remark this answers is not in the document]
coltraneyan: Yes! This solves so many issues I was trying to work around. Thank you.

PR #578, opened 2022-07-11
Feature/565 dropdown component

coltraneyan:
- New reusable dropdown component
- Styled according to this version below from Figma
178142351-ee911c42-8dd5-4dd6-9447-f2fa05f02a66

- in addition to options and current values, it also recieves a width prop to create a static width the parent can adjust according to the string length of the options
- all issues should be resolved from [this draft PR](https://github.com/airswap/airswap-web/pull/577) @piersss


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+          itemWidth={width}
+          onClick={() => {
+            handleShow();
+          }}
+        >
+          {options.map((option, index) => {
+            return (
+              <Item
+                isSelected={
+                  value
```
GiuntaLucas, line 57: You can optimise and just write value ? value.label === option.label : index === 0. Don't realy need to return true or false, the condition will do the job ^^

coltraneyan: Nice catch, thanks.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+import Icon from "../Icon/Icon";
+import { Item, Wrapper, AbsoluteWrapper, Current } from "./Dropdown.styles";
+
+export type SelectOption = {
+  label: string;
+  value: string;
+};
+
+export type DropdownProps = {
+  value?: SelectOption;
```
piersss, line 12:
```suggestion
  value: SelectOption;
```

In the designs there's always a default state. So `value` should not be optional.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+  width?: number;
+};
+
+export const Dropdown: React.FC<DropdownProps> = ({
+  value,
+  options,
+  onChange,
+  width,
+}) => {
+  const [showOptions, setShowOptions] = useState<Boolean>(false);
```
piersss, line 24:
```suggestion
```

We're using `focus` so this is also not needed.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+                  handleChange(option);
+                }}
+              >
+                {option.label}
+              </Item>
+            );
+          })}
+        </AbsoluteWrapper>
+      )}
+    </Wrapper>
```
piersss, line 75:
```suggestion
return (
    <Wrapper className={className}>
      <Current>
        {value.label}
        <Icon name={"chevron-down"} />
      </Current>
      <AbsoluteWrapper>
        {options
          .filter(option => option.value !== value.value)
          .map((option, index) => {
          return (
            <Item
              key={option.value}
              onClick={() => {
                handleChange(option);
              }}
            >
              {option.label}
            </Item>
          );
        })}
      </AbsoluteWrapper>
    </Wrapper>
```

In the end you just need a lot less. Added a filter to filter out the active option.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+  label: string;
+  value: string;
+};
+
+export type DropdownProps = {
+  value?: SelectOption;
+  options: SelectOption[];
+  onChange: (option: SelectOption) => void;
+  width?: number;
+};
```
piersss, line 16:
```suggestion
  className?: string;
};
```

Styled-components needs `className` to let a parent style it's children.

https://styled-components.com/docs/basics#styling-any-component

piersss (approved):

GiuntaLucas (approved):

PR #581, opened 2022-07-16
Focus/OnClick bug fix

coltraneyan:
-Fixed small bug which was giving :focus precedence over onClick, thus closing the dropdown before firing change functions
- onClick is now replaced by onMouseDown

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -38,7 +38,7 @@ export const Dropdown: React.FC<DropdownProps> = ({
             return (
               <Item
                 key={option.value}
-                onClick={() => {
+                onMouseDown={() => {
                   handleChange(option);
                 }}
               >


src/components/Dropdown/Dropdown.tsx
```diff
@@ -38,7 +38,7 @@ export const Dropdown: React.FC<DropdownProps> = ({
             return (
               <Item
                 key={option.value}
-                onClick={() => {
+                onMouseDown={() => {
```
piersss, line 41:
```suggestion
                `onPointerDown={() => {`
```
`onMouseDown` probably won't work on touch devices. https://caniuse.com/pointer

piersss (changes requested):

piersss:
I've added `onPointerDown` to my branch here: https://github.com/airswap/airswap-web/pull/582 So this one can be closed.

PR #583, opened 2022-07-16
Feature/576 expiry info

### LINKED ISSUE #576: Expiry info
Screenshot 2022-07-10 at 17 34 25

We can reuse the `getTimeBetweenTwoDates` helper from @codyenokida . Just need to refactor it a bit because the "ago" translations is baked into it.

Edit: really need date-fns for this.

piersss:
Fixes #576 

- Added new ExpiryIndicator component
- Added date-fns to project for ExpiryIndicator tooltip
- Refactored `getTimeBetweenTwoDates.ts` to make use of date-fns
- Added Pill and Tooltip styled-component

Screenshot 2022-07-16 at 16 27 19

### DIFF
--- a/src/components/ExpiryIndicator/ExpiryIndicator.tsx
+++ b/src/components/ExpiryIndicator/ExpiryIndicator.tsx
@@ -1,7 +1,8 @@
 import React, { FC } from "react";
 import { useTranslation } from "react-i18next";
 
-import { Container, Label, Strong } from "./ExpiryIndicator.styles";
+import { Pill } from "../../styled-components/Pill/Pill";
+import { Strong } from "./ExpiryIndicator.styles";
 
 type ExpiryIndicatorProps = {
   expiry: Date;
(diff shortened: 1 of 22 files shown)


src/components/ExpiryIndicator/ExpiryIndicator.tsx
```diff
@@ -0,0 +1,41 @@
...
+    [expiry, t]
+  );
+
+  return (
+    <Container>
+      <StyledPill>
+        {hasPassed ? t("orders.expired") : t("orders.expiresIn")}
+        {!hasPassed && <Strong>{timeLeft}</Strong>}
+      </StyledPill>
+      <StyledTooltip>{format(expiry, "dd-MMM-yyyy pppp")}</StyledTooltip>
```
GiuntaLucas, line 36: We should display the right format according to the local date of the user. 
I don't no the lib you are using here, but we can do that easely in Javascript with .localeDateString() for the date format + .localeTimeString() for the time format. A concat of both will give your a nice format according of the locale setting from the user. What do you think ?

GiuntaLucas (changes requested):

src/components/ExpiryIndicator/ExpiryIndicator.tsx
piersss: Good point. Just tried using localeDateString and localeTimeString, it works but I have to write an extra helper for getting the timezone info readable. So the best way is still date-fns I think.

GiuntaLucas: Yes if you prefer the date-fns libs go a head, i just don't no this lib, i'm use to work with native date object or moment.
If you can display to the user according to his local timezone, it would be verry nice for the user.

piersss: Date-fns is very similar to moment. So it also takes user's timezone into account.

PR #606, opened 2022-09-01
Feature/603 taker routing

### LINKED ISSUE #603: Catch faulty order strings and redirect to 404 page.
Compiling an order string should be done in the store. User should be redirected to OrderDetailWidget or a 404 order not found page.

Design will come up with 404 page.

piersss:
Fixes #603 

I also split the otc store into a make-otc and take-otc store.

@coltraneyan There's an `activeOrder` in the take-otc store you can use in the OrderDetailWidget. You can import it by using `useAppSelector` like so:

`const { activeOrder } = useAppSelector(selectTakeOtcReducer);`

Or you could pass it as a prop in `OrderDetail` to the `OrderDetailWidget`.


src/pages/OrderDetail/OrderDetail.tsx
```diff
@@ -1,11 +1,37 @@
...
 
 const OrderDetail: FC = () => {
+  const dispatch = useAppDispatch();
+  const { compressedOrder } = useParams<{ compressedOrder: string }>();
+
+  const { status } = useAppSelector(selectTakeOtcReducer);
+
+  useEffect(() => {
+    if (compressedOrder) {
+      dispatch(decompressAndSetActiveOrder(compressedOrder));
```
coltraneyan, line 19:
```suggestion
      dispatch<any>(decompressAndSetActiveOrder(compressedOrder));
```

probably a better fix, but I get a type/overload error here and I used this to stop it

coltraneyan (changes requested):

src/pages/OrderDetail/OrderDetail.tsx
```diff
@@ -1,11 +1,37 @@
...
+    return (
+      <Page>
+        <InvalidOrder />
+      </Page>
+    );
+  }
+
   return (
     <Page>
       <OrderDetailWidget />
```
coltraneyan, line 37: can now take in ```status``` and ```order``` props

src/pages/OrderDetail/OrderDetail.tsx
piersss: Strange, I don't get an error. Will add to fix things on your end.

src/pages/OrderDetail/OrderDetail.tsx
piersss: Cool!

coltraneyan (approved):

PR #614, opened 2022-09-25
Feature/611 no wallet support

### LINKED ISSUE #611: Take screen: show order even if wallet is not connected
Order should be at least viewable even if user has no wallet connected.

piersss:
Fixes #611


coltraneyan (approved):
Looks good! ```OrderDetailWidget/subcomponents/ActionButtons/ActionButtons.tsx``` needs one fix from my PR. 

buttonsDisabled should be something like this to allow for all unconnected wallet scenarios: 
```` 
const buttonDisabled =
    (hasInsufficientBalance || !isIntendedRecipient) &&
    !isExpired &&
    !isMakerOfSwap &&
    !isNotConnected;

PR #624, opened 2022-10-03
Feature/617 Taken State from Nonce Used

coltraneyan:
- added getOrderStatus hook
- added getTakenState which calls NonceUsed
- currently only displays status if user is on correct chain, other wise shows open state (library errors otherwise)

### DIFF
--- a/src/components/OrderDetailWidget/hooks/useOrderStatus.tsx
+++ b/src/components/OrderDetailWidget/hooks/useOrderStatus.tsx
@@ -4,7 +4,7 @@ import { FullOrder } from "@airswap/typescript";
 
 import { ethers } from "ethers";
 
-import { getTakenState } from "../../../features/takeOtc/takeOtcActions";
+import { getTakenState } from "../../../features/takeOtc/takeOtcHelpers";
 import { OrderStatus } from "../../../types/orderStatus";
 
 export const useOrderStatus = (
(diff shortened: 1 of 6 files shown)


src/features/takeOtc/takeOtcActions.ts
```diff
@@ -74,3 +77,17 @@ export const takeOtcOrder = createAsyncThunk(
...
+) => {
+  const SwapContract = new Contract(
+    swapDeploys[order.chainId],
+    SwapInterface,
+    library.getSigner()
+  );
+
+  let tx = await SwapContract.nonceUsed(order.signerWallet, order.nonce);
+  return tx;
+};
```
piersss, line 93: Actions are for `async` functions that mutate the store. This one doesn't mutate store so it should be a helper.

src/components/OrderDetailWidget/hooks/useOrderStatus.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [isTaken, setIsTaken] = useState(OrderStatus.open);
+
+  useMemo(() => {
+    if (library && chainId?.toString() === order.chainId) {
+      try {
+        getTakenState(order, library).then((r) =>
+          setIsTaken(r ? OrderStatus.taken : OrderStatus.open)
+        );
+      } catch (e) {
+        console.log(e);
```
piersss, line 24: `console.error(e)`

Any errors expected here?

piersss (changes requested):
Hooks!

PR #639, opened 2022-10-27
Fixes #638 token amount bug

### LINKED ISSUE #638: Token Amounts do not update for taker side
Screen Shot 2022-10-27 at 11 29 49 AM

coltraneyan:
- `side` in `SwapInputs` only updates token names, amounts now switch with `userIsMakerOfSwap` truthiness


piersss:
Nice find. But it means `side` is broken so you should fix that instead.

piersss (approved):

PR #642, opened 2022-10-30
375: Get protocol fee from contract

### LINKED ISSUE #375: Fetch `protocolFee` from SwapContract and use instead of hard coded values
Currently `1.0007` is used in a few places for last look calculations, and text around the interface says 0.07%. We should use the onchain value for future proofing.

piersss:
Fixes #375 

Also deleted tests that failed today, probably because today is 25 hours instead of 24.


coltraneyan (approved):
looks good!

PR #662, opened 2022-11-06
652: Added tooltip to order status indicator

### LINKED ISSUE #652: "Make" home screen: Add tooltip on status lights reading "Active" or "Inactive"

piersss:
Fixes #652

Also gave canceled orders a red indicator that has "Canceled" tooltip.

Screenshot 2022-11-06 at 21 36 23

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -65,7 +65,8 @@
     "fee": "Fee",
     "expiry": "Expiry",
     "pair": "Pair",
-    "close": "close"
+    "close": "close",
+    "active": "Active"
   },
   "information": {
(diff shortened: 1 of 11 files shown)


coltraneyan (changes requested):
found this bug on expired swaps: 
Screen Shot 2022-11-06 at 4 17 57 PM


in `OrderStatusInfo.tsx` line 23, something like this will fix it:

```
if (status === OrderStatus.open || OrderStatus.expired) {
    return <ExpiryIndicator expiry={expiry} className={className} />;
  }

coltraneyan (approved):

PR #665, opened 2022-11-08
Fix/653 dismiss order

### LINKED ISSUE #653: "Make" home screen: Update tooltip on dismiss order button to read "Dismiss"

piersss:
Fixes #653 

Also fixed "Tooltip does not dissappear after removing a swap from the list" bug.


coltraneyan (approved):
works nice!

PR #673, opened 2022-11-13
671: Added protocolFee to useShouldDepositNativeTokenAmount hook

### LINKED ISSUE #671: WETH: Required wrapping amount incorrect
From Twitter:
> Bug with amount of WETH, if we want to swap 0.11 ETH and already have 0.10WETH, dapp say "you need 0.01WETH more " but that it will wrap 0.11 ETH and not 0.01 ETH as indicated in the past message

piersss:
Fixes #671 

While testing the fix I got a SIGNER_BALANCE_LOW error. Figured out I had to add the 0.07% protocol fee. So I added that to the amount you have to wrap. Informed user in warning text.

Screenshot 2022-11-13 at 11 38 56

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -167,7 +167,7 @@
     "dismiss": "Dismiss",
     "youHaveNoOpenOTCOrders": "You have no open OTC orders",
     "startByCreatingANewOrder": "Start by creating a new order",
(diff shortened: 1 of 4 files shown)


coltraneyan:
It works well for me, the only issue is the changing of the window height. Could we use a toast or an overlay with a warning icon instead? Otherwise looks good. 
Screen Shot 2022-11-16 at 6 29 29 PM

coltraneyan (approved):

PR #677, opened 2022-11-18
675: Added max button to maker widget

### LINKED ISSUE #674: Signer fee should be prominent on Make Flow

### LINKED ISSUE #675: Add "Max" button to Make Flow
Max calculation should also factor in the signer fee.

piersss:
Fixes #674 
Fixes #675 

https://user-images.githubusercontent.com/86911296/202849061-68d21cee-5a8e-4e8b-b7da-a69f5ebb87f3.mov

One thing I noticed is that we round the amount with `stringToSignificantDecimals` so sometimes you're left with some dust amount which is a little annoying.

Also stumbled upon a bug with toAtomicString when trying to submit this dust amount. Created issue here:
https://github.com/airswap/airswap-protocols/issues/933

Edit: This error is intentional but we need to fix this in the interface.

### DIFF
--- a/src/components/MakeWidget/MakeWidget.tsx
+++ b/src/components/MakeWidget/MakeWidget.tsx
@@ -123,7 +123,7 @@ const MakeWidget: FC = () => {
     !makerAmount.length || parseFloat(makerAmount) === 0 || makerAmount === ".";
   const hasMissingTakerAmount =
     !takerAmount.length || parseFloat(takerAmount) === 0 || takerAmount === ".";
-  const maxAmount = useMaxAmount(makerAmount);
+  const maxAmount = useMaxAmount(makerTokenInfo?.address || null, true);
   const showMaxButton = !!maxAmount && makerAmount !== maxAmount;
   const showMaxInfoButton =
     !!maxAmount &&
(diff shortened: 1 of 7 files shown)


src/components/MakeWidget/MakeWidget.tsx
```diff
@@ -204,8 +204,8 @@ const MakeWidget: FC = () => {
         ? wrappedTokenAddresses[chainId!]
         : takerTokenAddress;
 
-    const formattedMakerAmount = stringToSignificantDecimals(makerAmount, 4);
-    const formattedTakerAmount = stringToSignificantDecimals(takerAmount, 4);
+    const formattedMakerAmount = stringToSignificantDecimals(makerAmount, 6);
+    const formattedTakerAmount = stringToSignificantDecimals(takerAmount, 6);
```
coltraneyan, line 208: like you said, dust was lowered for me when createOrder takes in a non rounded makerAmount. could move stringToSignificantDecimals out of the setState.

piersss: Yes I tried that but then the `toAtomicString` could fail give an `fractional component exceeds decimals` errors. This is because the smart contract's `signerAmount` has to be an integer.

coltraneyan: Ah I understand now. Oops!

coltraneyan (approved):

Issue #679, opened 2022-11-23
Add tooltip and focus ring to counterparty input field, when taker address not valid


2xAST:
https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/airswap-web?node-id=13189%3A15515&t=2hoAzepk26s6r1p4-4

Same tooltip as in https://github.com/airswap/airswap-web/issues/678

See also: https://github.com/airswap/airswap-web/issues/667#issuecomment-1310558246

piersss:
Assigned this to helloitsme3

PR #691, opened 2022-12-07
Added tooltip and focus ring when address is invalid

helloitsm3:
PR for #679

### DIFF
--- a/src/components/MakeWidget/MakeWidget.styles.tsx
+++ b/src/components/MakeWidget/MakeWidget.styles.tsx
@@ -74,7 +74,8 @@ export const StyledInfoSection = styled(InfoSection)`
 export const StyledTooltip = styled(Tooltip)`
   position: absolute;
   right: 0;
-  top: -0.7rem;
+  top: 0;
+  height: 3.5rem;
 `;
 
 export const TooltipContainer = styled.div`
--- a/src/components/MakeWidget/MakeWidget.styles.tsx
+++ b/src/components/MakeWidget/MakeWidget.styles.tsx
@@ -2,6 +2,7 @@ import styled from "styled-components/macro";
 
 import breakPoints from "../../style/breakpoints";
 import { SelectLabel } from "../../styled-components/Select/Select";
+import Tooltip from "../SwapInputs/subcomponents/Tooltip/Tooltip";
 import ActionButtons from "./subcomponents/ActionButtons/ActionButtons";
 import AddressInput from "./subcomponents/AddressInput/AddressInput";
 import InfoSection from "./subcomponents/InfoSection/InfoSection";
@@ -68,4 +69,16 @@ export const StyledInfoSection = styled(InfoSection)`
   & + ${StyledActionButtons} {
     margin-top: 1rem;
   }
+`;
+
+export const StyledTooltip = styled(Tooltip)`
+  position: absolute;
+  right: 0;
+  top: -0.7rem;
+`;
+
+export const TooltipContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  position: relative;
 `;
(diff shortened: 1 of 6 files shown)


src/components/MakeWidget/MakeWidget.tsx
```diff
@@ -152,6 +154,10 @@ const MakeWidget: FC = () => {
     useShouldDepositNativeToken(makerAmount);
   const shouldDepositNativeToken = !!shouldDepositNativeTokenAmount;
   const hasDepositPending = useDepositPending();
+  const inputHasError =
```
piersss, line 157:
```suggestion
  const addressInputHasError =
```

piersss (changes requested):
1. Error state shows immediately when user enters character. This is not user friendly behaviour.
2. Border of input should be red, not any div's around it.

src/components/MakeWidget/MakeWidget.tsx
```diff
@@ -310,6 +314,18 @@ const MakeWidget: FC = () => {
     }
   };
 
+  const handleTakerAddress = async (value: string) => {
```
GiuntaLucas, line 317: Same here, i think we should use a method return type.

GiuntaLucas (changes requested):

src/components/MakeWidget/MakeWidget.tsx
helloitsm3: we no longer need this. I'm removing it

src/components/MakeWidget/MakeWidget.styles.tsx
```diff
@@ -69,3 +70,15 @@ export const StyledInfoSection = styled(InfoSection)`
     margin-top: 1rem;
   }
 `;
+
+export const StyledTooltip = styled(Tooltip)`
+  position: absolute;
+  right: 0;
+  top: -0.7rem;
```
piersss, line 77:
```suggestion
  top: 0;
  height: 3.5rem;
```

src/components/MakeWidget/MakeWidget.tsx
```diff
@@ -357,11 +360,23 @@ const MakeWidget: FC = () => {
...
-          onInfoButtonClick={toggleShowOrderTypeInfo}
-        />
+        <TooltipContainer>
+          <StyledAddressInput
+            hasError={!!error}
+            value={takerAddress}
+            onChange={(e) => {
+              setTakerAddress(e);
+              dispatch(reset());
+            }}
```
piersss, line 370:
```
const handleAddressInputChange = (value: string) => {
    setTakerAddress(value);
    if (error?.type === AppErrorType.invalidAddress) {
      dispatch(setError(undefined))
    }
  }
```

src/components/SwapInputs/subcomponents/Tooltip/Tooltip.styles.tsx
```diff
@@ -8,13 +8,14 @@ const triangleSize = 14;
...
-  width: 100%;
   max-width: 17rem;
-  height: 4.5rem;
+  height: 0;
 
   @media ${breakPoints.tabletLandscapeUp} {
     align-items: center;
     top: 0;
+    width: 100%;
+    height: 4.5rem;
```
piersss, line 18: Oh I see that this styling was broken, nice!

piersss (changes requested):
import errors:

```
src/components/MakeWidget/MakeWidget.tsx
  Line 13:10:  'ethers' is defined but never used     @typescript-eslint/no-unused-vars
  Line 56:10:  'ErrorList' is defined but never used  @typescript-eslint/no-unused-vars
```

Input border wrong on focus:
Screenshot 2022-12-14 at 22 12 42

Disable Take button when error.

piersss (approved):

Issue #688, opened 2022-12-05
Ratefield styling improvements


piersss:

piersss:
So in this scenario amount should be shortened to 10199. Then there's enough space for "USDT" and "ETH" to appear.

PR #692, opened 2022-12-07
Added tooltip when hovering rate field

helloitsm3:
#688

### DIFF
--- a/src/components/MakeWidget/MakeWidget.tsx
+++ b/src/components/MakeWidget/MakeWidget.tsx
@@ -313,7 +313,7 @@ const MakeWidget: FC = () => {
     }
   };
 
-  const getRate = (): String => {
+  const getRate = (): string => {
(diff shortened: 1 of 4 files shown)


src/components/MakeWidget/MakeWidget.tsx
```diff
@@ -310,6 +313,16 @@ const MakeWidget: FC = () => {
     }
   };
 
+  const getRate = () => {
```
GiuntaLucas, line 316: You should add a method return type to be consistant

GiuntaLucas (changes requested):

piersss (changes requested):
1. Uhhh... I dont think you actually tested your code.

Screenshot 2022-12-12 at 10 17 00
Screenshot 2022-12-12 at 10 16 47
Screenshot 2022-12-12 at 10 33 38

2. Make sure to do mobile styling too.

Issue #703, opened 2022-12-19
Align copy text in order toasts


coltraneyan:
some toasts messages have periods, some do not. quick overview of all toast copy to make sure it matches a single standard

PR #705, opened 2022-12-22
Fixes #702 #703 OTC Toasts

### LINKED ISSUE #702: Fix My Order Cancel Toast
dismiss order shows validator error, as it calls the cancel function. we want this order to be dismissed as it has already been cancelled and this is an unnecessary error message and function call.

coltraneyan:
- calls `getNonceUsed` before attempting to cancel on chain to avoid unnecessary error messages
- fixes minor period adjustments to keep copy consistent in toasts


piersss (approved):
LGTM

PR #706, opened 2022-12-23
WIP: 694: Allow swaps to be made with arbitrary tokens by address

### LINKED ISSUE #694: Allow swaps to be made with arbitrary tokens by address
When selecting a token in "Make" mode, user should be able to provide an arbitrary contract address. Once provided, the app should attempt to read ERC20 properties from that contract and enable the user to create a swap from or to the token.

One concern is scams and duplicate tokens. Therefore, once selected, on the "Make" or the "Swap" screens, a token not found in metadata should have a warning icon next to it to indicate that it is unknown and should be verified.

piersss:
Fixes #694 

**Scenario 1, maker flow:**

https://user-images.githubusercontent.com/86911296/209365319-fd1576f1-1454-4030-a4b8-91d8349909ce.mov

1. Newly scraped tokens will be added to `all` tokens in metadata store.
2. After you import it it will be added to `active` and `custom` in metadata.

**Scenario 2, taker flow:**

Screenshot 2022-12-23 at 17 13 10

1. Modified `useTakerTokenInfo` hook for this. If token is not found in metadata store, then add to `all` tokens in metadata store.
2. Since the token is not `active` it will appear as "Inactive" token in the user's TokenList

**Scenario 3: unconnected wallet taker flow:**

Screenshot 2022-12-25 at 22 30 41

When wallet is not connected `scrapeToken` can't be called because it needs a provider. If there's also no cached token that matches address then a "Unknown token, please connect" appears. Might need a design for this but we can do that in different ticket.

**Scenario 4: swap flow:**

Screenshot 2022-12-23 at 17 17 23

1. The user is unable to use the custom token since it's not active.

I also did some refactoring of caching of metadata (moved it all too `metadataSubscriber`) hopefully I didn't break stuff and this PR is getting too long.


coltraneyan:
I have no token info when wallet not connected - Swaps is actually quite clean:
Screen Shot 2022-12-26 at 5 34 47 PM

but with viewing OTC swaps its an issue:
Screen Shot 2022-12-26 at 5 34 56 PM

I experimented a bit with the provider, try using REACT_APP_RPC's in the env as the providers when we have no library.

piersss:
Fixed coltrane's issue by adding the rpc providers in .env. `scrapeToken` can now be called without a provider. Although it sometimes faild because of the `429` error: `{"detail":"Request was throttled. Expected available in 1 second."}`. But let's make a separate ticket for that.

coltraneyan (approved):
nice! looks good on my end, didn't run into the 429 error.

piersss:
Thanks. 429 errors seems only to be happening on test net, so that's nice.

PR #717, opened 2023-01-05
Firefox on windows icon not showing properly

filipmanole:
The solution I found was to Increase the font size to separate the dot from 'i'.

### DIFF
--- a/src/components/TokenSelect/TokenSelect.styles.tsx
+++ b/src/components/TokenSelect/TokenSelect.styles.tsx
@@ -73,7 +73,7 @@ export const MaxButtonStyle = css`
   align-self: center;
   border-radius: 0.125rem;
   font-weight: 600;
-  font-size: 0.75rem;
+  font-size: 0.81rem;
   line-height: 1;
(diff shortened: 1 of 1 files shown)


filipmanole:
On Windows, the letter 'i' is not centred perfectly as on Mac.

src/components/TokenSelect/TokenSelect.styles.tsx
```diff
@@ -73,7 +73,7 @@ export const MaxButtonStyle = css`
   align-self: center;
   border-radius: 0.125rem;
   font-weight: 600;
-  font-size: 0.75rem;
+  font-size: 0.81rem;
```
piersss, line 76: https://github.com/airswap/airswap-web/blob/develop/CONTRIBUTING.md#styled-components

> We define dimensions using rem based on 16px

filipmanole:
The font is not supported on windows. The solution was to change it to a web-safe font.

piersss (approved):
LGTM

PR #727, opened 2023-01-27
725: Improved clearing store when switching between wallets

### LINKED ISSUE #725: State issue switching provider networks with same chainId
User reports on `airswap-web-dev` env that "the frontend seems to cache the token list; when I switch back to local node it still has the token list (but can't connect to my account anymore)"

Image

piersss:
Fixes #725 

- Improved cleanup of store when switching / disconnecting wallet
- TokenList will close when disconnected
- Fixed wallet button indicator when wallet is not connected.

### DIFF
--- a/src/components/SwapWidget/SwapWidget.tsx
+++ b/src/components/SwapWidget/SwapWidget.tsx
@@ -245,6 +245,12 @@ const SwapWidget: FC = () => {
     [quoteAmount]
   );
 
+  useEffect(() => {
+    if (!active) {
+      setShowTokenSelectModalFor(null);
+    }
+  }, [active]);
+
   const hasSufficientAllowance = (tokenAddress: string | undefined) => {
     if (tokenAddress === nativeCurrency[chainId || 1].address) return true;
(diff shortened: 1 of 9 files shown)


src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -758,7 +758,7 @@ const SwapWidget: FC = () => {
 
       <Overlay
         onCloseButtonClick={() => setShowTokenSelectModalFor(null)}
-        isHidden={!showTokenSelectModalFor}
+        isHidden={!showTokenSelectModalFor || !active}
```
coltraneyan, line 761: for both Make and Swap, this is cleaner when the user reconnects, as the tokenModal re-opens in a clunky manner in current setup

```
useEffect(() => {
    if (!active) {
      setShowTokenSelectModalFor(null)
    }
  }, [active])```

piersss: Thanks. Fixed!

coltraneyan (approved):
🚀

Issue #556, opened 2022-06-28
Available Swaps Screen (Taker)


dmosites:
https://www.figma.com/file/DTJwC2mQ9kvyjM0171tqHR/airswap-web?node-id=10341%3A7930

PR #736, opened 2023-02-15
Indexer and IndexerRegistry work

### LINKED ISSUE #573: Implement Indexer API Client

gpxl-dev:
This closes #573 and #572, paving the way for #556.

Adds the `indexerSlice` which is not currently used anywhere, but provides the groundwork.
At this time I haven't implemented a call to get _all_ orders on the servers as I'm not sure that's warranted, but assume you can achieve this anyway by simply passing a non-restrictive filter.


piersss:
Looks good but it's not working for me. I'm getting this error:

```[indexerSlice] Order request failed for http://airswap.mitsi.ovh/ Method does not exist.```

dmosites:
> Looks good but it's not working for me. I'm getting this error:
> 
> `[indexerSlice] Order request failed for http://airswap.mitsi.ovh/ Method does not exist.`

This appears to be a server error, which is okay. We cannot rely on Indexer servers being available. Only if they are and we get orders back are those orders considered in the taker flow.

dmosites (approved):

PR #743, opened 2023-03-19
740: Added fallback and loading state to custom tokens when wallet is not connected

### LINKED ISSUE #740: Orders with custom tokens do no get displayed when wallet is not connected
It will be displayed after connecting the wallet.

Image

piersss:
Fixes #740 

https://user-images.githubusercontent.com/86911296/226141601-c0399ef6-3fdd-43dc-b60a-ae00756c30e3.mov


coltraneyan (approved):
lgtm!

PR #744, opened 2023-03-20
upgrade @airswap/libraries to 4.0.4

dmosites:


piersss (approved):
Tested swap and otc, all still work.

PR #754, opened 2023-04-18
latest beta packages; new library api; tooling upgrades

dmosites:
Using new libraries for Registry, Server, SwapERC20, etc.
Using webpack 5 via react-scripts upgrade.
Dependency upgrades and cleanup.
Tested on Goerli with RFQ and OTC swaps.


src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -134,12 +134,10 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
...
-    const errors = await check(
-      order,
-      order.senderWallet,
-      order.chainId,
-      library!.getSigner()
-    );
+    let errors: any[] = [];
+    if (library) {
+      errors = await check(order, order.senderWallet, order.chainId, library);
+    }
```
piersss, line 140:
```suggestion
    if (!library) {
      return;
    }

    const errors = await check(
      order,
      order.senderWallet,
      order.chainId,
      library,
    );
```

src/components/Toasts/TransactionToast.tsx
```diff
@@ -41,15 +41,15 @@ export type TransactionToastProps = {
   /**
    * Token Info of sender token
    */
-  senderToken?: TokenInfo;
+  senderToken?: TokenInfo | null;
```
piersss, line 44: Don't do null and undefined together.

```suggestion
  senderToken?: TokenInfo;
```

In ToastController:
```
<TransactionToast
            onClose={() => toast.dismiss(t.id)}
            type={type}
            transaction={transaction}
            senderToken={senderToken || undefined}
            signerToken={signerToken || undefined}
            error={error}
          />
```

piersss:
OTC swaps work but RFQ not, are the makers empty?

dmosites:
> OTC swaps work but RFQ not, are the makers empty?

It takes some seconds for the test maker to spin up after idling so might try a couple times. Just worked on my end.

piersss (approved):
Tested again and all works.

PR #769, opened 2023-05-20
766: Added fix for ReviewApprovalInfo and approve token

### LINKED ISSUE #766: Fix order completed screen OTC
- Transaction Completed screen should only show up for taker

piersss:
Fixes #766 

Also added fix for approval amount not checking for fee.

### DIFF
--- a/src/components/MakeWidget/MakeWidget.tsx
+++ b/src/components/MakeWidget/MakeWidget.tsx
@@ -1,4 +1,4 @@
-import { FC, useContext, useEffect, useState } from "react";
+import { FC, useContext, useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useHistory } from "react-router-dom";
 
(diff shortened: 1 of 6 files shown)


personnamedmike:
When I go to Make an order and enter an amount in ETH that starts with a decimal, for example `.1` I see the error below. The error does not happen if I enter a number before the decimal, for example `0.1`. This is happening on mainnet

Screenshot 2023-05-20 at 12 54 54 PM
Screenshot 2023-05-20 at 12 55 44 PM

personnamedmike (approved):
I looked through all the code. It looks good to me, but it might be wise to get a 3rd set of eyes on it from someone more familiar with the code base. 

I left a comment about a bug that I found. Not sure if it's directly related to these code changes.

personnamedmike:
re: bug.
Adding `type="number"` to line 159 in TokenSelect.tsx seems to fix this issue

> When I go to Make an order and enter an amount in ETH that starts with a decimal, for example `.1` I see the error below. The error does not happen if I enter a number before the decimal, for example `0.1`. This is happening on mainnet
> 
> Screenshot 2023-05-20 at 12 54 54 PM Screenshot 2023-05-20 at 12 55 44 PM

Issue #731, opened 2023-02-03
Uncaught error causing white screen


agriimony:
On https://airswap-web-dev.herokuapp.com/#/make 
Type "." into ETH amount in the "From" field.
Whole screen turns white.

On console:
`Uncaught Error: missing value (argument="value", value=".", code=INVALID_ARGUMENT, version=bignumber/5.7.0)`

personnamedmike:
I opened a PR to fix this issue https://github.com/airswap/airswap-web/pull/771

personnamedmike:
I'm going to close this issue
cc @piersss

PR #771, opened 2023-05-20
added type="number" seems to fix the issue

personnamedmike:
This is a quick fix in regards to issue #731


piersss:
I'm not getting a white screen on staging. So pretty sure this was already fixed.

personnamedmike:
Ok, it must only be on the `develop` branch then. I'll close the issue

PR #780, opened 2023-05-23
773: Fixed translations

### LINKED ISSUE #773: Requested updates to language and controls
My Orders
* canceled order status should be “CANCELED” rather than expiry time
* “Make Order” → “Make New Order”

Make
* “Specific Wallet” should be “Specify Taker” and button should read “Enter Taker Address”

Swap (Maker)
* buttons should be “Make New Order” and “Cancel Order”
* “Cancel swap” → “Cancel order”
* With a canceled order, secondary button should be “Back”

Swap (Taker)
* “​​This order is not for the connected wallet” → “This order is for a different taker”
* “New Swap” button should go to Taker flow.

When wrapping ETH
* Body message: "You'll need to wrap at least 1.0007 ETH to make this order."

piersss:
Fixes #773 

Also updated all translations in poeditor.


.env.example
```diff
@@ -1,8 +1,8 @@
 # RPC URLs (for WalletConnect and @airswap/metadata to work properly)
 # Mainnet
-REACT_APP_RPC_URL_1=""
-# Rinkeby
-REACT_APP_RPC_URL_5=""
+REACT_APP_RPC_URL_1=https://mainnet.infura.io/v3/40797ab9860349d3aa5abf747ca5db0d
```
personnamedmike, line 3: Should these keys be exposed like this on GitHub?

public/locales/en/translation.json
```diff
@@ -127,13 +126,13 @@
...
+    "makeNewOrder": "Make new order",
     "moreInformation": "More information",
     "nativeCurrencyMaxInfoText": "{{amount}}. This amount is calculated after subtracting a {{fee}} gas fee from your total balance.",
     "newQuoteIn": "New Quote In",
-    "newSwap": "New swap",
-    "noIndexersFound": "No indexer nodes found",
     "noIndexerOrdersFound": "No orders found for this token pair",
-    "orderSuccessfullyCreated": "Order successfully created",
+    "noIndexersFound": "No indexer nodes found",
+    "orderSuccessfullyCreated": "Order successfully created.",
```
personnamedmike, line 135: this line has a period at the end but most other lines don't. Will that create inconsistent UX?

personnamedmike (approved):
Looks fine, but double check env.example file. Might need to change pk's

piersss:
> Looks fine, but double check env.example file. Might need to change pk's

Yes, deleted it. Oops!

PR #781, opened 2023-05-24
fix/751 OTC Swap Flow

coltraneyan:
- addresses all issues from feedback in https://github.com/airswap/airswap-web/pull/755 
- uses redux to handle searchAmount state with correct cleanup
- adds AvailableSwapsWidget to OrderDetailWidget as well for smoother transitions

* all server querying is the same, however, "Aggregate Error" can happen when re-querying. Not sure if this will be fixed by Mitsi and Whatdays current work etc. but lets please tackle this in another PR.

### DIFF
--- a/src/components/OrderDetailWidget/OrderDetailWidget.tsx
+++ b/src/components/OrderDetailWidget/OrderDetailWidget.tsx
@@ -144,6 +144,8 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
     return new Date(parseInt(order.expiry) * 1000);
   }, [order]);
 
+  const userIsFromSwapFlow = !!location.state?.fromSwapFlow;
+
   const [showFeeInfo, toggleShowFeeInfo] = useToggle(false);
   const [showViewAllQuotes, toggleShowViewAllQuotes] = useToggle(false);
 
@@ -151,7 +153,8 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
     if (!indexerUrls && library) {
       dispatch(fetchIndexerUrls({ provider: library }));
     }
-  }, [dispatch, indexerUrls, library]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [indexerUrls, library]);
 
   useEffect(() => {
     if (indexerUrls) {
@@ -166,22 +169,20 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
         })
       );
     }
-  }, [dispatch, indexerUrls, senderToken?.address, signerToken?.address]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [indexerUrls, senderToken?.address, signerToken?.address]);
 
-  // swap flow handlers
-  const userIsFromSwapFlow = !!location.state?.fromSwapFlow;
-
-  const backToViewAllQuotes = () => {
+  // button handlers
+  const backToSwapPage = () => {
     history.push({
       pathname: `/${AppRoutes.swap}/${senderToken?.address}/${signerToken?.address}`,
       state: { fromSwapFlow: true },
     });
   };
 
-  // button handlers
   const handleBackButtonClick = () => {
     if (userIsFromSwapFlow) {
-      backToViewAllQuotes();
+      backToSwapPage();
       return;
     }
     history.push({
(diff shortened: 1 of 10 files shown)


src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -119,9 +131,45 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
...
+
+  useEffect(() => {
+    if (!indexerUrls && library) {
+      dispatch(fetchIndexerUrls({ provider: library }));
+    }
+  }, [dispatch, indexerUrls, library]);
+
+  useEffect(() => {
+    if (indexerUrls) {
+      dispatch(
```
piersss, line 144: This is called 2 times when arriving from SwapWidget. You should probably only have indexers as dependency.

src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -119,9 +131,45 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
   }, [order]);
 
   const [showFeeInfo, toggleShowFeeInfo] = useToggle(false);
+  const [showViewAllQuotes, toggleShowViewAllQuotes] = useToggle(false);
+
+  useEffect(() => {
+    if (!indexerUrls && library) {
+      dispatch(fetchIndexerUrls({ provider: library }));
+    }
+  }, [dispatch, indexerUrls, library]);
```
piersss, line 140:
```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [indexerUrls, library]);
```

src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -119,9 +131,45 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
...
+            offset: 0,
+            limit: 100,
+          },
+        })
+      );
+    }
+  }, [dispatch, indexerUrls, senderToken?.address, signerToken?.address]);
+
+  // swap flow handlers
+  const userIsFromSwapFlow = !!location.state?.fromSwapFlow;
```
piersss, line 158: This should be higher with the variable declarations. Not between useEffect's.

src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -119,9 +131,45 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
...
+
+  const backToViewAllQuotes = () => {
```
piersss, line 160:
```suggestion
  const backToSwapPage = () => {
```

src/components/Overlay/Overlay.tsx
```diff
@@ -31,6 +31,7 @@ export type OverlayProps = {
    * Hide or show the component
    */
   isHidden?: boolean;
+  animate?: boolean;
```
piersss, line 34:
```suggestion
  shouldAnimate?: boolean;
```

piersss (changes requested):
This works a lot better. Adding AvailableOrdersWidget to OrderDetailWidget is a good move.

Note: the translation is changed from `orders.availableSwaps` to `orders.availableOrders`.

Small bug when routing directly to swap from order detail widget:

https://github.com/airswap/airswap-web/assets/86911296/53a403d5-8fcb-4345-9936-f898cb342c57

piersss (changes requested):
Why don't you use tradeTerms.baseAmount instead of indexer.currentSearchAmount?

Please rebase your branch to develop / merge develop into your branch.

src/components/OrderDetailWidget/OrderDetailWidget.tsx
```diff
@@ -118,10 +130,47 @@ const OrderDetailWidget: FC<OrderDetailWidgetProps> = ({ order }) => {
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [indexerUrls, library]);
+
+  useEffect(() => {
+    if (indexerUrls && senderToken?.address && signerToken?.address) {
+      dispatch(
+        getFilteredOrders({
+          filter: {
+            senderTokens: [senderToken.address],
```
piersss, line 150: To get orders with similar amounts we should add `senderMinAmount` and `senderMaxAmount` -25% + 25% respectively.

coltraneyan: i had this implemented before and we decided to take it out. if we want to do this, lets move it to another PR.

coltraneyan:
> Why don't you use tradeTerms.baseAmount instead of indexer.currentSearchAmount?
> 
> Please rebase your branch to develop / merge develop into your branch.

i'm pretty sure the components cleanup functions will clear the trade terms. current search amount is just essentially a pinned value we need for a very specific point in time, which i think is best to have its own state rather than piggyback on another.

piersss (approved):
LGTM

PR #783, opened 2023-05-26
Fix/762 - (***REVIEW BRANCH "Fixes #762". IGNORE THIS PR)

### LINKED ISSUE #762: Add ability to target a specific server URL via query string
When set, query from a specific server URL rather than using the Registry.
May need design to indicate the currently targeted server URL.

personnamedmike:
Fixes #762

To test this feature on localhost, enter the following URL:
`https://localhost:3000/#/swap/:senderToken/:signerToken?serverURL=server.com`

For `server.com` in the URL: on Goerli you can use the server https://airswap-goerli-http.herokuapp.com/.
A working example URL on Goerli is: http://localhost:3000/#/swap/native/0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6?serverURL=https://airswap-goerli-http.herokuapp.com/

Some possible servers on mainnet for `server.com` are:
https://airswap-pmm-rfq-server-i24uhfiu3fh.alphalab.cc
https://b2c2.xyz/airswap/
https://aomcfsa7.altono.xy
https://ast.ngrok.io

A working example URL on mainnet is: http://localhost:3000/#/swap/native/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48?serverURL=https://aomcfsa7.altono.xyz

Note that the translations for `order.selectedServer` are blank except for English


public/locales/dk/translation.json
```diff
@@ -100,6 +100,7 @@
         "tokenPairUnavailable": "",
         "retryOrCancel": "",
         "newSwap": "",
+        "selectedServer": "",
```
piersss, line 103: When developing only edit `en/translation.json`. Translation team will take care of the other files. Also it will automatically fallback to english if undefined, so that's perfectly fine.

public/locales/en/translation.json
```diff
@@ -147,6 +147,7 @@
     "revertPrice": "Revert price",
     "scanningPeers": "Scanning peers on the network",
     "searchByNameOrAddress": "Search name or paste address",
+    "selectedServer": "Querying the selected server...",
```
piersss, line 150:
```suggestion
    "selectedServer": "Selected {{serverUrl}} as custom server",
```

`Querying the selected server...` suggests it's in a loading state. But design will take a look at this to communicate this better.

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -165,6 +168,14 @@ const SwapWidget: FC = () => {
...
   const [isWrapping, setIsWrapping] = useState(false);
   const [isRequestingQuotes, setIsRequestingQuotes] = useState(false);
+  // if `bestTradeOption` is true, `isQueryingSelectedServer` will get toggled to false. This is so InfoSection.tsx renders the correct text & data
+  const [isQueryingSelectedServer, setIsQueryingSelectedServer] =
+    useState(false);
+
+  // Check if URL query string contains server URL
+  const location = useLocation();
+  const query = new URLSearchParams(location.search);
+  const queryUrl = query.get("serverURL");
```
piersss, line 178: Make a custom `useSearchParams` hook for this so you end up with this:

```suggestion
const serverUrl = useSearchParams<{ serverUrl: string }>();
```

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -710,6 +654,98 @@ const SwapWidget: FC = () => {
...
+  }, [baseTokenInfo, dispatch, indexerUrls, quoteTokenInfo]);
+
+  // check if serverURL query param exists
+  useEffect(() => {
+    if (queryUrl) {
+      setIsQueryingSelectedServer(true);
+      setServerUrl(queryUrl);
+    } else {
+      setIsQueryingSelectedServer(false);
+      setServerUrl(null);
```
piersss, line 736: This will clear the serverUrl when order changes. Is this supposed to happen?

src/components/SwapWidget/subcomponents/InfoSection/InfoSection.tsx
```diff
@@ -22,6 +22,7 @@ import {
 export type InfoSectionProps = {
   isApproving: boolean;
   isConnected: boolean;
+  isSelectedServer: boolean;
```
piersss, line 25:
```suggestion
  hasSelectedCustomServer: boolean;
```

src/components/SwapWidget/subcomponents/InfoSection/InfoSection.tsx
```diff
@@ -116,6 +118,16 @@ const InfoSection: FC<InfoSectionProps> = ({
     );
   }
 
+  // logic below renders 'Querying the selected server...' into <InfoSection>
```
piersss, line 121:
```suggestion
```

Unnecessary explanation

piersss (changes requested):
I think the custom serverUrl should stay for the entire session. Don't clear when changing tokens.

You suggest using this url:

http://localhost:3002/#/swap/native/0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6?serverURL=https://airswap-goerli-http.herokuapp.com/

But this is a ETH to WETH trade, so that will wrap the ETH, so no server is used for that. Did you test with a real token pair?

I tested it with USDC -> USDT and that seems to work. 👍

public/locales/dk/translation.json
personnamedmike: Ok, will remove everything from non en files

public/locales/en/translation.json
personnamedmike: I committed this change, but `{{serverUrl}}` isn't pulling in any dynamic variables

src/components/SwapWidget/SwapWidget.tsx
personnamedmike: I'll paste in the code because I made changes and the lines in the file are now different.

```if (queryUrl) {
      setIsQueryingSelectedServer(true);
      setServerUrl(queryUrl);
    } else {
      setIsQueryingSelectedServer(false);
      setServerUrl(null);
    }```

line 735 sets the value to false and line 736 to null because it's the `else` conditional to the previous block, which reads `if (queryURL) {`

File: src/components/SwapWidget/SwapWidget.tsx (hunk 3 above)
Comment (personnamedmike) [reply src/components/SwapWidget/SwapWidget.tsx]: I made a new file with the hook `useSearchParams`. Pushed it to this branch

File: src/components/SwapWidget/subcomponents/InfoSection/InfoSection.tsx (hunk 5 above)
Comment (personnamedmike) [reply src/components/SwapWidget/subcomponents/InfoSection/InfoSection.tsx]: Added in change and pushed to this branch

**Comment by personnamedmike** [comment]
I'm unable to test out anything on Goerli. No token pairs seem to exist with ETH, other than WETH (which isn't a pair)

File: public/locales/en/translation.json (hunk 2 above)
Comment (personnamedmike) [reply public/locales/en/translation.json]: I fixed this up, but the issue is that the dynamic variable comes out as a URL. I'm not sure if servers have names that are nice to read. Originally @dmosites mentioned to say `"Querying the selected server..."`. How should we proceed with this?

Screenshot 2023-06-01 at 4 45 49 PM

**Comment by personnamedmike** [comment]
@piersss @dmosites I'm unable to test this out correctly on Goerli. The only pair that works in ETH - WETH. ETH - USDC, or ETH - USDT doesn't work. 

On Mainnet I'm able to test out ETH - USDC on the server below, but unless I click "complete swap" I'm not able to test further. I don't want to test with real funds: http://localhost:3000/#/swap/native/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48/?serverURL=https://ast.ngrok.io/.

This is a blocker. How should I proceed?

File: src/components/SwapWidget/SwapWidget.tsx (hunk 3 above)
Comment (piersss) [reply src/components/SwapWidget/SwapWidget.tsx]: Now the dispatch in `useSearchParams` runs every render. Please install redux dev tools so you can catch this kind of mistakes.

Again, please make a `useSearchParams` that's reusable, so you can use it like this, with a custom generic:

`const serverUrl = useSearchParams<{ serverUrl: string }>();`

File: src/components/SwapWidget/SwapWidget.tsx [hunk 7]
```diff
@@ -710,6 +669,83 @@ const SwapWidget: FC = () => {
     }
   };
 
+  useEffect(() => {
```
piersss, line 672: What's the actual change here?

src/components/SwapWidget/SwapWidget.tsx
```diff
@@ -307,36 +244,51 @@ const SwapWidget: FC = () => {
...
+    value: string,
+    serverURL: string | null
+  ) => {
     const baseRoute = `/${AppRoutes.swap}`;
     const tokenPairs = getTokenPairs(type, value, quoteToken, baseToken);
     const tokenFrom = transformAddressAliasToAddress(tokenPairs.tokenFrom!);
     const tokenTo = transformAddressAliasToAddress(tokenPairs.tokenTo!);
     const tokenFromAlias = transformAddressToAddressAlias(tokenFrom);
     const tokenToAlias = transformAddressToAddressAlias(tokenTo);
+    const queryString = `?serverURL=${serverURL}`;
```
piersss, line 258: Don't store in url, since you're already saving it to the store.

src/components/SwapWidget/SwapWidget.tsx
personnamedmike: What should the purpose of `useSearchParams` be? I'm trying to use it as a hook that return the query string from the URL. If the hook takes in a string as an argument, then it won't return a query string. It's purpose would then be to call dispatch and set the query string into state in Redux. If the hook calls dispatch, then it will be continuously called in the SwapWidget component. Not sure I'm understanding this

src/components/SwapWidget/SwapWidget.tsx
personnamedmike: I'm not sure about this useEffect hook. I never worked on it. I moved some code around, so my name might appear next to it on Git

personnamedmike:
Please view PR #786 instead of this. That PR is a copy of this but with more updated code from the latest merged branch

PR #787, opened 2023-06-12
Fixes#770 - Add a link to block explorer on Order Completed screen

### LINKED ISSUE #770: Add a link to block explorer on Order Completed screen

personnamedmike:
Fixes #770 

https://github.com/airswap/airswap-web/assets/86911296/d5ec16b8-439a-4fb8-97f0-464ca7d96df6


piersss:
Right now you assume that the last transaction in the store is the transaction of the current swap. A user could have multiple transactions running at the same time. So it's best not to assume and make sure you have the right transaction.

Is it possible to compare the data from `bestTradeOption` and find the right order that way? It either returns a `rfqOrder` or an `lastLookOrder`. `rfqOrder` has an `orderErc20` which has a `nonce`. You can compare the `nonce` with the order in transaction.

`lastLookOrder` does not have an `orderErc20` so that makes things a bit complicated... maybe just use `quoteAmount` then.

personnamedmike:
> quoteAmount

@piersss I fixed useTransactionHash.ts so it takes in a nonce from `bestTradeOption`, then checks to see if it's in the list of all transactions. Let me know if this is what you meant

piersss (approved):
LGTM

personnamedmike:
Feature demo:

https://github.com/airswap/airswap-web/assets/90465420/764a0016-1ef1-48a6-be9d-dc0bdd1601c5

PR #788, opened 2023-06-19
Fixes732 - Add a link to etherscan contract in token list

personnamedmike:
https://github.com/airswap/airswap-web/assets/90465420/74817c79-48ae-4dd6-80bd-f8211619ec44


personnamedmike:
Waiting on confirmation from design team that we're going with this design

personnamedmike:
Fixed up styling and added tooltip

https://github.com/airswap/airswap-web/assets/90465420/ad51c71e-c8af-4652-a531-15bd39f350c2

public/locales/en/translation.json
```diff
@@ -63,6 +63,7 @@
     "token": "Token",
     "undefined": "undefined",
     "unknown": "Unknown",
+    "verifyEtherscan": "Verify on Etherscan",
```
dmosites, line 66: Each chain has a different block explorer, so how about just "Verify token"?

dmosites (changes requested):

public/locales/en/translation.json
personnamedmike: fixed

dmosites (approved):

PR #790, opened 2023-06-24
if no selectedToken, animation renders

personnamedmike:


personnamedmike:
Feature is no longer needed

PR #823, opened 2023-07-24
Fix/114 - SwapWidget: after completing a swap the allowance is not updated

personnamedmike:
https://github.com/airswap/airswap-web/assets/90465420/3ae9975b-4dc8-4788-ad20-fc3644d3fa86


piersss (approved):
LGTM

PR #846, opened 2023-09-04
Fix `toMaxAllowedDecimalsNumberString`

gpxl-dev:
This PR fixes [a problem an OTC maker was having](https://discord.com/channels/590643190281928738/599315389327081492/1147978174253236366) whereby they couldn't sign their order.

They have tried to enter a number of decimals that is unsupported by the tokens, so the parse is failing.

The UI does already have a check function for this `toMaxAllowedDecimalsNumberString`, but for some reason this function ignores numbers > 1.

I've updated the function & test.


piersss:
`toMaxAllowedDecimalsNumberString` is not where the rounding should be because the creating the order we add the fee amount. Based on that total amount we round to the max amount of decimals. That happens in `MakeOrderReview` `toRoundedNumberString`.

piersss:
I'll have fix tomorrow:


https://github.com/airswap/airswap-web/assets/86911296/7e9d1230-2f5e-4892-a79d-70f81cb20e9e

piersss (approved):
Approving because you can't approve your own PR.

gpxl-dev:
lgtm!

PR #886, opened 2024-03-16
853: Clicking "Back" during pending transaction should reset

### LINKED ISSUE #853: RFQ, OTC: Clicking "Back" during pending transaction should reset
Clicking "Back" during pending transaction (e.g. Approve) currently returns to previous screen.
Since the transaction has already started at this point, it's safe to reset the query.

piersss:
Fixes #853 

https://github.com/airswap/airswap-web/assets/86911296/cf2643ea-1f3e-4c3f-bfd5-6c83bd7e2638

Fixed for Make, Take and RFQ widget.

When clicking back from approve or wrap it takes you back to the start screen. Removed any loading state in start screen so user is free to go back to approve or wrap screen or change the query.

Resetting the query is not necessary I think but could still add that if necessary.

### DIFF
--- a/src/components/@reviewScreens/ApproveReview/ApproveReview.tsx
+++ b/src/components/@reviewScreens/ApproveReview/ApproveReview.tsx
@@ -27,6 +27,7 @@ import {
 } from "./ApproveReview.styles";
 
 interface ApproveReviewProps {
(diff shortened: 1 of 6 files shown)


dmosites (approved):
LGTM

Issue #900, opened 2024-04-25
Fix gas store api


piersss:
```
GET https://ethgasstation.info/api/ethgasAPI.json net::ERR_FAILED 404 (Not Found)
getFastGasPrice @ gasCostApi.ts:20
gasCostApi.ts:24 Error getting gas price from ethgas.watch API:  Failed to fetch
```

Fix the gas cost api so we can properly compare RFQ vs LastLook with the current gas price.

PR #901, opened 2024-04-29
899: Added quotes store for LastLook and RFQ pricing

### LINKED ISSUE #899: Add quotes store for RFQ and LastLook orders
Right now there's a big function called `requestQuotes` in SwapWidget that handles the fetching of RFQ and LastLook orders with promises. To make it more maintainable and readable we need to move this logic to it's own store and a nice clean `createAsyncThunk`.

Also LastLook data is handled via `createContext` which is confusing because all the other data is handled via `redux`. So hopefully I can remove the whole `LastLookProvider`.

ghost:
Fixes #899 

- Moved all logic for RFQ and LastLook pricing in the quotes store and useQuotes hook.
- Added a debug mode, type "debug" and you can disable RFQ or LastLook for better testing.
- Added logging when in debug mode.
- Cleanup of legacy code
- Discovered that gas api is broken which is needed for LastLook order calculation, will fix next: #900 

https://github.com/airswap/airswap-web/assets/86911296/92ab906d-040e-49b9-911f-d1e9f2812a65

https://github.com/airswap/airswap-web/assets/86911296/43553ed3-38a6-4845-b6ca-82407f9c9155

### DIFF
--- a/src/components/@widgets/SwapWidget/SwapWidget.tsx
+++ b/src/components/@widgets/SwapWidget/SwapWidget.tsx
@@ -426,7 +426,6 @@ const SwapWidget: FC = () => {
   const handleActionButtonClick = async (action: ButtonActions) => {
     switch (action) {
       case ButtonActions.goBack:
-        console.log("goback", state);
         setState(SwapWidgetState.overview);
         dispatch(clearQuotes());
         dispatch(clear());


dmosites (approved):
LGTM

PR #916, opened 2024-06-05
915: Revert swap contract to v4

### LINKED ISSUE #915: Revert swap contract to 4.0

piersss:
Fixes #915

### DIFF
--- a/src/helpers/swapErc20.ts
+++ b/src/helpers/swapErc20.ts
@@ -573,7 +573,7 @@ export const getSwapErc20Contract = (
   providerOrSigner: ethers.providers.Provider | ethers.Signer,
   chainId: number
 ): ethers.Contract => {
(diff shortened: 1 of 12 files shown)


src/helpers/swapErc20.ts
```diff
@@ -0,0 +1,592 @@
...
+    stateMutability: "nonpayable",
+    type: "function",
+  },
+];
+
+export const getSwapErc20Contract = (
+  providerOrSigner: ethers.providers.Provider | ethers.Signer,
+  chainId: number
+): ethers.Contract => {
+  if (!shouldOverride || chainId !== 1) {
```
dmosites, line 576: Let's check whether `chainId` is in `mainnets` from `@airswap/constants` so that all prod chains would be overridden.

mikestarrdev (approved):

PR #945, opened 2024-07-31
update otc for v4 compatibility

dmosites:


src/features/orders/ordersActions.ts
```diff
@@ -277,6 +277,19 @@ export const approve =
...
 
+      if (isAppError(tx)) {
+        const appError = tx;
+        dispatch(setErrors([appError]));
+
+        if (appError.error && "message" in appError.error) {
+          dispatch(declineTransaction(appError.error.message));
+        }
+
+        dispatch(setStatus("failed"));
```
piersss, line 288:
```suggestion
        dispatch(setStatus("failed"));
        handleOrderError(dispatch, tx);
```

This does the same basically.

src/features/orders/ordersHelpers.ts
```diff
@@ -203,7 +209,7 @@ export async function check(
   provider: ethers.providers.Web3Provider,
   isSwapWithWrap?: boolean
 ): Promise<AppError[]> {
-  const strings = await (
+  const [count, strings] = await (
```
piersss, line 212: strings is now undefined so it breaks here

src/features/orders/ordersActions.ts
dmosites: Great. Had just copied this over from the function below, will update.

src/features/orders/ordersHelpers.ts
dmosites: Works on my end and the swapErc20 contractV4Abi ABI indicates uint256, bytes32[]:
```
name: "check",
outputs: [
  { internalType: "uint256", name: "", type: "uint256" },
  { internalType: "bytes32[]", name: "", type: "bytes32[]" },
],
```

piersss (approved):
LGTM

PR #1036, opened 2025-08-31
Fix: Added overwrite limit order warning

makkelie-dev:
Thing I noticed that's needed while testing.

- When calculating total approve amount of all limit orders, don't include orders with the pair of the new orders because it will be overwritten.
- Added warning when overwriting limit order

Screenshot 2025-08-31 at 15 34 58


src/components/ApprovalNotice/helpers.ts
```diff
@@ -128,6 +129,17 @@ const filterTokenOrder = async (
     return false;
   }
 
+  console.log(takerTokenAddress, isDelegateRule(order) && order.signerToken);
```
dmosites: console.log

dmosites (approved):

PR #1046, opened 2025-09-08
Fix: native currency max amount

### LINKED ISSUE #1041: Fix Max Button and Native Currency Safe Transaction Fee Logic
**Description:**
When a user’s balance is less than 0.01 (the hardcoded nativeCurrencySafeTransactionFee), the app automatically subtracts this fee, resulting in a displayed balance of 0.

The user then sees a message:
> “0 ETH. This amount is calculated after subtracting a 0.01 gas fee from your total balance.”

This is confusing because:
- The Max button still appears but only fills in 0.
- The fixed 0.01 fee may no longer be relevant as transaction fees have dropped significantly.

**Expected Behavior:**
- The Max button should not be shown if the resulting max value is 0.
- The safe transaction fee should not be a fixed 0.01. Instead, it should be dynamically calculated based on current gas price estimates.

**Proposed Fix:**
- Hide the Max button when the post-fee amount is 0.
- Replace the hardcoded 0.01 nativeCurrencySafeTransactionFee with a dynamic calculation using live gas price data.
- Update messaging to clearly explain the calculation in case of insufficient funds.

Image

makkelie-dev:
Fixes #1041 

- Hide the Max button when the max amount is 0 after subtracting gas fees.
- Replace the hardcoded 0.01 gas fee with the dynamic calculation from the gas price store.
- Update messaging to clearly explain the calculation in cases of insufficient funds.
- Fixed styling of the information button.

Screenshot 2025-09-08 at 21 37 36


dmosites (approved):
If possible please make the "info" icon a lowercase "i" for clarity.

